PR #15, opened 2019-12-12
added support for reverse relations

wouterdb:
This add support to the language server to support not only jump to definition but also jump to usage


sanderr:
I am missing some context. I digged a little in the code, but most methods have no type annotations and no docstrings, so I can't really comment on the logic. Apart from that everything looks good. Should I add another reviewer before approving?

PR #29, opened 2020-10-27
Issue/18 ls error reporting

### LINKED ISSUE #18: Add error reporting to language-server
capture whatever exceptions from the compiler and report via 'squigly red line' 

see https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_publishDiagnostics

sanderr:
closes #18

### DIFF
--- a/server/src/inmantals/server.py
+++ b/server/src/inmantals/server.py
@@ -26,7 +26,7 @@ from inmantals.jsonrpc import (
 import os
 import types
 from asyncio import Lock
-from typing import Optional
+from typing import Dict, Optional
 from inmanta import compiler
 from inmanta.util import groupby
 from intervaltree.intervaltree import IntervalTree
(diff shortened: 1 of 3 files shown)


server/src/inmantals/server.py
```diff
@@ -109,11 +122,33 @@ def treeify_reverse(iterator):
...
+                    )
+                await self.publish_diagnostics(params)
+            logger.exception("Compile failed")
+
         except Exception:
             logger.exception("Compile failed")
 
     async def initialized(self):
-        await asyncio.get_event_loop().run_in_executor(self.threadpool, self.compile_and_anchor)
+        await self.compile_and_anchor()
```
sanderr, line 151: @wouterdb I'm not sure about this change, wdyt?

server/src/inmantals/server.py
```diff
@@ -109,11 +122,33 @@ def treeify_reverse(iterator):
...
+                if error.location is None:
+                    params = None
+                else:
+                    params = lsp_types.PublishDiagnosticsParams(
+                        uri=error.location.uri,
+                        diagnostics=[
+                            lsp_types.Diagnostic(
+                                range=error.location.range,
+                                severity=lsp_types.DiagnosticSeverity.Error,
+                                message=f"{error.type}: {error.message}",
```
wouterdb, line 140: Is the type a relevant part of the message?

server/src/inmantals/server.py
wouterdb: 1. I'm not sure this is according to the language server spec
2. Why block the IDE side on this?

wouterdb (changes requested):

server/src/inmantals/server.py
sanderr: I find it useful but that might be because I've got insight in the compiler. It might be better to just stick with the error message.

server/src/inmantals/server.py
sanderr: Does this really block the IDE side? Cause this is all happening in the language server, right?
Anyway, the reason I made this change is because `compile_and_anchor` now does an async call to `send_notification`. I thought this might not be the right approach to deal with that but I'm not sure what would be. Do you have a suggestion?

wouterdb: https://docs.python.org/3/library/asyncio-task.html#creating-tasks

server/src/inmantals/server.py
wouterdb: (don't know either, tbh)

server/src/inmantals/server.py
wouterdb: Sorry, I'm wrong, this logically blocking already.

The reason for the threadpool is because compilation can take some time, blocking the IOLoop, making the whole lsp unresponsive. so you can move the executor call deeper into the structure

wouterdb (changes requested):
see comment

server/src/inmantals/server.py
sanderr: I've added a comment above the `run_in_executor` line. Could you verify that it describes the motivation for using the executor correctly?

wouterdb: I think it is good

wouterdb (approved):

andraskvr (approved):

PR #30, opened 2020-10-27
Create virtual env automatically

### LINKED ISSUE #23: Ensure venv is automatically created when the plugin is enabled.
This refers to the venv the language server is installed in, which must be manually specified and created at the moment.

andraskvr:
Closes #23

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,7 +5,7 @@ import * as net from 'net';
 import * as cp from 'child_process';
 import * as fs from 'fs';
 
-import { workspace, ExtensionContext, Disposable, window } from 'vscode';
+import { workspace, ExtensionContext, Disposable, window, Uri } from 'vscode';
 import { RevealOutputChannelOn, LanguageClient, LanguageClientOptions, ServerOptions, TransportKind, Executable, ErrorHandler, Message, ErrorAction, CloseAction } from 'vscode-languageclient';
 
 export function activate(context: ExtensionContext) {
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -150,10 +145,35 @@ export function activate(context: ExtensionContext) {
...
+		}
+		const venvDir = Uri.joinPath(context.globalStorageUri, ".env").fsPath;
+		const venvPath = Uri.joinPath(context.globalStorageUri, ".env", "bin", "python3").fsPath;
+
+		if (!fs.existsSync(venvDir)) {
+			const venvProcess = cp.spawnSync("python3", ["-m", "venv", venvDir ]);
+			if (venvProcess.status != 0) {
+				window.showErrorMessage(`Virtual env creation at ${venvDir} failed with code ${venvProcess.status}, ${venvProcess.stderr}`);
+			}
+			const child = cp.spawnSync(venvPath, ["-m", "pip", "install", "--pre", "inmantals"]);
```
sanderr, line 164: Why `--pre`?

sanderr (changes requested):

src/extension.ts
sanderr: I see now it was there before your changes as well.

andraskvr: Yes, but now that you mention it, I'm not sure if it's necessary

src/extension.ts
```diff
@@ -150,10 +145,35 @@ export function activate(context: ExtensionContext) {
 
 	const enable: boolean = workspace.getConfiguration('inmanta').ls.enabled
 
+	function createVenvIfNotExists() {
+		const pp: string = workspace.getConfiguration('inmanta').pythonPath;
+		if (pp && fs.existsSync(pp)) {
+			window.showInformationMessage(`Using Python Interpreter at ${pp}`);
```
wouterdb, line 151: This message will pop up every time the extensions starts?

src/extension.ts
wouterdb: good point, should probably be removed, those who want --pre can do it by hand

wouterdb: (perhaps both pip-install's have to go into one method)

wouterdb (changes requested):

sanderr (approved):

wouterdb (approved):

PR #33, opened 2020-10-28
Issue/17 compile test for typescript part of extension

edvgui:
### Current progress:
1. Compile a correct basic model
- [x] Compile actually happen
- [x] The tests checks that the lib and env folders are created
- [x] The tests checks that no error occurred during compile

2. ~~Do a CTRL + Click~~ --> https://github.com/inmanta/vscode-inmanta/issues/38
3. ~~Verify that installing a venv works~~ --> https://github.com/inmanta/vscode-inmanta/issues/39


edvgui:
How could I check that the compilation went as expected? And is there a way to check if it has finished?

package.json
```diff
@@ -64,6 +64,9 @@
       }
     }
   },
+  "mocha": {
+    "exit": true
```
andraskvr, line 68: I think this is not required now

server/src/inmantals/server.py
```diff
@@ -255,7 +258,15 @@ async def dispatch_method(self, id, method, params):
 
         return result
 
-    async def send_show_message(self, type, message):
+    async def send_show_message(self, type: int, message: str):
+        """
+        Show a pop up message to the user, type gives the level of the message, message is the content
+        Available types are:
+         * 1: Error
```
andraskvr, line 265: Maybe add an Enum?

src/extension.ts
```diff
@@ -24,91 +24,98 @@ export function activate(context: ExtensionContext) {
...
 				}
-			)
+			);
 		}
 
 		diagnose() {
-			if (this._child != undefined)
-				return
+			if (this._child !== undefined)
+				{return;}
```
andraskvr, line 77:
```suggestion
			if (this._child !== undefined) {
				return;
			}
```

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	beforeEach((done) => {
+		Promise.all([
+			fs.writeFile(logPath, ""),
+			fs.remove(libsPath),
+			fs.remove(envPath),
+			fs.remove(modelUri.fsPath),
+		]).then(async values => {
+			await commands.executeCommand('workbench.action.closeActiveEditor');
+			done();
```
andraskvr, line 57: If you chain a `then()` to the `executeCommand` and call the `done` callback there, can't the explicit waiting in the test case be removed?

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+				// Opening model file
+				const doc: TextDocument = await workspace.openTextDocument(modelUri);
+				const edit: TextEditor = await window.showTextDocument(doc);
+
+				// Making file dirty and save it
+				const position: Position = new Position(0, 0);
+				const snippet: SnippetString = new SnippetString("\n");
+				await edit.insertSnippet(snippet, position);
+				assert.strictEqual(doc.isDirty, true, "The file should be dirty, but isn't");
+				await doc.save();
```
andraskvr, line 80: You have to both close the editor and make changes to the file? Isn't one of them enough to trigger the recompile?

andraskvr (changes requested):

server/src/inmantals/server.py
sanderr: (`inmantals.lsp_types` is the place to add it)

server/src/inmantals/server.py
```diff
@@ -144,11 +145,13 @@ def treeify_reverse(iterator):
                     ],
                 )
             await self.publish_diagnostics(params)
-            logger.exception("Compile failed")
+            await self.send_show_message(1, "Compilation failed: " + e.get_message())
```
sanderr, line 148: If `params` is set, vscode shows the error location in the file and lists it under "problems" (the diagnostics call). Showing a message in that case would be somewhat obsolete and it might draw attention away from the more informative diagnostics. On the other hand it would be consistent. I'm not sure which would be best. @andraskvr wdyt?
```suggestion
            if params is None:
                await self.send_show_message(1, "Compilation failed: " + e.get_message())
```

src/test/suite/compile.test.ts
edvgui: If I just change the file, I get an error with vscode because there is a conflict between what has been changed in the file manually (using `fs`) and the change I try to appy.  And if I don't change the file, the file is never dirty, so I need both yes.

package.json
edvgui: Indeed

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+	const start = Date.now();
+	return new Promise<boolean>((resolve, reject) => {
+		const readLogInterval = setInterval(() => {
+			if (Date.now() - start > timeout) {
+				reject(new Error("Timeout reached"));
+			} else {
+				fs.ensureFileSync(logPath);
+				fs.readFile(logPath, 'utf-8', (err, data) => {
+					if (err) {
+						console.log(err);
```
sanderr, line 25: Call reject? It's been a while since I've worked with javascript so I'm a bit out of sync.

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+				await fs.copyFile(source, modelUri.fsPath);
+				
+				// Wait one second to let vscode notice we closed the previous editor
+				await new Promise(resolve => setTimeout(resolve, 1000));
+
+				// Opening model file
+				const doc: TextDocument = await workspace.openTextDocument(modelUri);
+				const edit: TextEditor = await window.showTextDocument(doc);
+
+				// Making file dirty and save it
```
sanderr, line 75: I don't even think it's required for the file to be dirty

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import { after, before, describe, it, beforeEach } from 'mocha';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+
+import { Uri, window, commands, workspace, TextDocument, TextEditor, Position, SnippetString } from 'vscode';
+
+const logPath: string = '/tmp/vscode-inmanta.log';
+const workspaceUri: Uri = Uri.file(path.resolve(__dirname, '../../../src/test/workspace'));
+const libsPath: string = path.resolve(workspaceUri.fsPath, 'libs');
+const envPath: string = path.resolve(workspaceUri.fsPath, '.env');
```
sanderr, line 11: This will probably have to change once you merge in master

src/test/suite/compile.test.ts
```diff
@@ -0,0 +1,106 @@
...
+				await doc.save();
+
+				const succeeded = await waitForCompile(10000);
+				assert.strictEqual(succeeded, test.succeed);
+
+				const libsExists = fs.pathExistsSync(libsPath);
+				assert.strictEqual(libsExists, true, "The libs folder hasn't been created");
+
+				const envExists = fs.pathExistsSync(envPath);
+				assert.strictEqual(envExists, true, "The .env folder hasn't been created");
```
sanderr, line 89: The directory is not necessarily `.env`. Either describe it or use `envPath` in the message.

sanderr (changes requested):

src/test/suite/compile.test.ts
edvgui: No because the `executeCommand` actually gives an argument, and done will refuse any call with an argument.

server/src/inmantals/server.py
edvgui: Did it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/939de709d3b0da790beb077a77b621e46ae04848

package.json
edvgui: Changed it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/939de709d3b0da790beb077a77b621e46ae04848

src/test/suite/compile.test.ts
edvgui: Actually it is, or at least, it is the only way I found to make everything work

src/test/suite/compile.test.ts
edvgui: Why? Does the place where the environment is created changes?

sanderr: Indeed, see #30

edvgui: From https://github.com/inmanta/vscode-inmanta/pull/33#discussion_r515010393 I guess it does, I will take a look.

andraskvr:
I'm not sure where to put this, but the tests should be run in the CI pipeline, so `npm run test` or something similar should be added to the Jenkinsfile

server/src/inmantals/lsp_types.py
```diff
@@ -53,6 +53,18 @@ class DiagnosticSeverity(Enum):
     Hint: int = 4
 
 
+class MessageType(Enum):
+    """
+    Message type.
+    `LSP spec 3.15 <https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#window_showMessage>`__
```
sanderr, line 59:
```suggestion
```
I've moved this to a general spec reference at the top of the document (in the current master).

sanderr (changes requested):

src/test/suite/compile.test.ts
edvgui: I don't see where/when/how the environment folder is created, how is the location decided? Is there an environment variable set somewhere?

edvgui:
> I'm not sure where to put this, but the tests should run in the CI pipeline, so `npm run test` or something similar should be added to the Jenkinsfile

@andraskvr  Yes it should, I will try that, but won't we have some problems with Jenkins? The test might need a screen as it is opening a vscode window, no?

src/test/suite/compile.test.ts
edvgui: Not sure about that, if for some reason the read operation fails, it doesn't mean it couldn't succeed later on (before the timeout expire).  Like for example if the file doesn't exist yet, or is temporarily unreachable.

server/src/inmantals/server.py
edvgui: Ideally I would have liked for the message to be clickable, and to redirect toward the "problems" panel, but I couldn't find a way.

andraskvr:
> > I'm not sure where to put this, but the tests should run in the CI pipeline, so `npm run test` or something similar should be added to the Jenkinsfile
> 
> @andraskvr Yes it should, I will try that, but won't we have some problems with Jenkins? The test might need a screen as it is opening a vscode window, no?

I think it should be possible to run them in headless mode (there are frontend tests running in headless mode on Jenkins already). You can try something like `xvfb-run npm test`

src/test/suite/compile.test.ts
edvgui: Found it, I didn't merge yet

src/test/suite/compile.test.ts
sanderr: Right

src/test/suite/compile.test.ts
edvgui: Is it really the same venv as for `compilerVenv` though? `.env` still gets created for me, no matter the path I set for the venv.

sanderr: Hmm, it shouldn't, right @andraskvr?

andraskvr: That's right.
I checked out this branch and the tests indeed fail on this point for me.
Note that the compiler venv change affected both the language server and the extension. So you have to update both for it to work (`pip install -U server/` or something similar, and `vsce package`, then Extensions -> Install from VSIX for the extension part)

edvgui: I "fixed it" here https://github.com/inmanta/vscode-inmanta/pull/33/commits/920a3825eb58b9276f22b9d15b0df69f25eb4e2b, but it doesn't work for me.  
A part of it is that I don't know how to get the default path (inside the tests I mean, I would need to access the context of the extension, but I couldn't figure it out yet), which will be in my case: `/home/guillaume/.config/Code/User/workspaceStorage/6607cd65b9884be7da6505e5121d968a/Inmanta.inmanta/.env-ls-compiler`.
Another part is that that `.env` folder keeps getting created for me anyway.

edvgui: Actually, I simply removed all checks on the venv, as this as to be verified in this issue right? https://github.com/inmanta/vscode-inmanta/issues/39

Jenkinsfile
```diff
@@ -22,6 +22,7 @@ pipeline {
                 dir("server"){
                     sh "$INMANTA_TEST_ENV/bin/python3 -m tox --recreate"
                 }
+                sh 'rm -rf node_modules; npm i --also=dev; xvfb-run npm run test'
```
sanderr, line 25: Perhaps separate client and server tests in different stages?

edvgui: Good idea, I will do that

edvgui: Did it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/decabc3b9573bc0da5f374725efb20e2621c1a7b

edvgui:
> > > I'm not sure where to put this, but the tests should run in the CI pipeline, so `npm run test` or something similar should be added to the Jenkinsfile
> > 
> > 
> > @andraskvr Yes it should, I will try that, but won't we have some problems with Jenkins? The test might need a screen as it is opening a vscode window, no?
> 
> I think it should be possible to run them in headless mode (there are frontend tests running in headless mode on Jenkins already). You can try something like `xvfb-run npm test`

Finally managed to do it! https://github.com/inmanta/vscode-inmanta/pull/33/commits/aadc412f055606006b1f5b58bf14a5240bd03f96

Jenkinsfile
```diff
@@ -13,22 +13,32 @@ pipeline {
 
     environment {
       INMANTA_TEST_ENV="${env.WORKSPACE}/env"
+      INMANTA_SERVER_PATH="${env.WORKSPACE}/server"
+      INMANTA_PYTHON_PATH="${env.WORKSPACE}/env/bin/python3"
```
sanderr, line 17: similar here:
```suggestion
      INMANTA_LS_PYTHON_PATH="${env.WORKSPACE}/env/bin/python3"
```

src/test/runTest.ts
```diff
@@ -0,0 +1,49 @@
...
+	try {
+		// Loading config of testing workspace
+		const pythonPath = process.env.INMANTA_PYTHON_PATH;
+		const compilerVenv = process.env.INMANTA_COMPILER_VENV;
+		if (pythonPath === undefined) {
+			throw new Error("INMANTA_PYTHON_PATH has to be set");
+		}
+		if (compilerVenv === undefined) {
+			throw new Error("INMANTA_COMPILER_VENV has to be set");
+		}
```
sanderr, line 17: Why make this configurable?

sanderr (changes requested):

Jenkinsfile
```diff
@@ -13,22 +13,32 @@ pipeline {
 
     environment {
       INMANTA_TEST_ENV="${env.WORKSPACE}/env"
+      INMANTA_SERVER_PATH="${env.WORKSPACE}/server"
```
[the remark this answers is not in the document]
sanderr: Also, it looks like this is not used anywhere, it it?

src/test/runTest.ts
edvgui: I thought that it might be useful for the test on venv, but given that i am not testing it here I should probably remove it

sanderr: I think that would be best, the person who picks up the ticket can then decide if it's required.

edvgui: Removed it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/861a5702ef2483206fe5e16bfd83e9a51b630d47

Jenkinsfile
sanderr: (similar to https://github.com/inmanta/vscode-inmanta/pull/33#discussion_r515816250)

Jenkinsfile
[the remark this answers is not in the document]
edvgui: It is used when installing the language server in the venv used for the extension testing, but I agree that the name is poorly chosen, I will update that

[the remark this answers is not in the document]
sanderr: You're right, I found it now.

Jenkinsfile
edvgui: For that one I am less sure, I was referring to the setting: `inmanta.pythonPath`

sanderr: Which is the python path of the environment where the inmanta language server is (or should be) installed.

edvgui: Is it better like this? https://github.com/inmanta/vscode-inmanta/pull/33/commits/7a93af6a1916cadbef688f0eaf529b7450e48659

Jenkinsfile
```diff
@@ -13,22 +13,32 @@ pipeline {
...
+                    sh "$INMANTA_PYTHON_PATH -m tox --recreate"
                 }
             }
         }
+        stage('Extension tests') {
+            steps {
+                sh 'rm -rf node_modules'
+                sh 'npm i --also=dev'
+                sh '$INMANTA_PYTHON_PATH -m pip install -e $INMANTA_SERVER_PATH'
+                sh 'INMANTA_COMPILER_VENV="" xvfb-run npm run test'
```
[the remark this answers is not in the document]
edvgui: Did it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/7a93af6a1916cadbef688f0eaf529b7450e48659

Jenkinsfile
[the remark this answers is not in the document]
edvgui: Changed it in https://github.com/inmanta/vscode-inmanta/pull/33/commits/7a93af6a1916cadbef688f0eaf529b7450e48659

sanderr (approved):

andraskvr (approved):

PR #34, opened 2020-10-29
Issue/19 symbol provider

### LINKED ISSUE #19: add symbol provider
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol

sanderr:
closes #19

### DIFF
--- a/server/src/inmantals/server.py
+++ b/server/src/inmantals/server.py
@@ -110,6 +110,7 @@ class InmantaLSHandler(JsonRpcHandler):
             # fresh project
             Project.set(Project(self.rootPath))
 
+            # can't call compiler.anchormap and compiler.get_types_and_scopes directly because of inmanta/inmanta#2471
             compiler_instance: compiler.Compiler = compiler.Compiler()
             (statements, blocks) = compiler_instance.compile()
             scheduler_instance = scheduler.Scheduler()
--- a/server/src/inmantals/server.py
+++ b/server/src/inmantals/server.py
@@ -38,7 +38,6 @@ from concurrent.futures.thread import ThreadPoolExecutor
 from tornado.iostream import BaseIOStream
 from inmanta import resources
 from inmanta.agent import handler
-from inmanta.ast.attribute import Attribute
 from inmanta.ast.entity import Entity, EntityLike, Implementation
 from inmanta.execute import scheduler
 from inmanta.module import Project
--- a/server/src/inmantals/server.py
+++ b/server/src/inmantals/server.py
@@ -16,6 +16,7 @@
     Contact: code@inmanta.com
 """
 import asyncio
+import inspect
 import json
 import logging
 import os
@@ -64,6 +65,7 @@ class InmantaLSHandler(JsonRpcHandler):
         self.state_lock: asyncio.Lock = asyncio.Lock()
         self.diagnostics_cache: Optional[lsp_types.PublishDiagnosticsParams] = None
         self.supported_symbol_kinds: Optional[Set[lsp_types.SymbolKind]] = None
+        self.compiler_venv_path: Optional[str] = None
 
     async def initialize(self, rootPath, rootUri, capabilities: Dict[str, object], **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
@@ -71,6 +73,9 @@ class InmantaLSHandler(JsonRpcHandler):
         self.rootPath = rootPath
         self.rootUrl = rootUri
         os.chdir(rootPath)
+        init_options = kwargs.get("initializationOptions", None)
+        if init_options:
+            self.compiler_venv_path = init_options.get("compilerVenv", os.path.join(self.rootPath, ".env-ls-compiler"))
 
         value_set: List[int]
         try:
@@ -116,8 +121,13 @@ class InmantaLSHandler(JsonRpcHandler):
             resources.resource.reset()
             handler.Commander.reset()
 
+            project_signature = inspect.signature(Project.__init__)
             # fresh project
-            Project.set(Project(self.rootPath))
+            if "venv_path" in project_signature.parameters.keys() and self.compiler_venv_path:
+                logger.debug("Using venv path " + str(self.compiler_venv_path))
+                Project.set(Project(self.rootPath, venv_path=self.compiler_venv_path))
+            else:
+                Project.set(Project(self.rootPath))
 
             # can't call compiler.anchormap and compiler.get_types_and_scopes directly because of inmanta/inmanta#2471
             compiler_instance: compiler.Compiler = compiler.Compiler()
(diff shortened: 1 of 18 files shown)


server/src/inmantals/server.py
```diff
@@ -229,6 +251,54 @@ async def textDocument_references(self, textDocument, position, context):  # noq
...
+                return lsp_types.SymbolKind.Function
+            if isinstance(tp, inmanta_type.ConstraintType):
+                return lsp_types.SymbolKind.Class
+            if isinstance(tp, EntityLike):
+                return lsp_types.SymbolKind.Class
+            if isinstance(tp, Implementation):
+                return lsp_types.SymbolKind.Constructor
+
+            logger.exception("Unknown type %s, using default symbol kind" % tp)
+            return if_supported(lsp_types.SymbolKind.Object, lsp_types.SymbolKind.Variable)
```
sanderr, line 279: I tried my best to pick the `SymbolKind` that I found best matched the symbol type in the inmanta DSL but most of them are far from a perfect fit. Let me know if you think there are better matches possible.

server/src/inmantals/server.py
```diff
@@ -229,6 +251,54 @@ async def textDocument_references(self, textDocument, position, context):  # noq
...
+                name=name, kind=get_symbol_kind(tp), location=self.convert_location(tp.location)
+            )
+            for name, tp
+            in matching_types
+        )
+
+        attribute_symbols: Iterator[lsp_types.SymbolInformation] = (
+            lsp_types.SymbolInformation(
+                name=attribute_name,
+                kind=lsp_types.SymbolKind.Field,
```
sanderr, line 292: `Property` would be an option here as well. I haven't got a strong preference.

server/tests/test_smoke.py
```diff
@@ -165,3 +184,96 @@ async def test_connection(client, caplog):
...
+                ),
+            ),
+        ),
+        lsp_types.SymbolInformation(
+            name="testmodule::plugin_symbol_test",
+            kind=lsp_types.SymbolKind.Function,
+            location=lsp_types.Location(
+                uri=uri_testmodule_plugins,
+                range=lsp_types.Range(
+                    start=lsp_types.Position(line=4, character=0), end=lsp_types.Position(line=4, character=1)
```
arnaudsjs, line 264: This range doesn't seem to be correct, is it?

server/tests/test_smoke.py
```diff
@@ -165,3 +184,96 @@ async def test_connection(client, caplog):
...
+            ),
+        ),
+        lsp_types.SymbolInformation(
+            name="symbol",
+            kind=lsp_types.SymbolKind.Field,
+            location=lsp_types.Location(
+                uri=uri_testmodule_model,
+                range=lsp_types.Range(
+                    start=lsp_types.Position(line=1, character=0), end=lsp_types.Position(line=1, character=0)
+                ),
```
arnaudsjs, line 275: This range doesn't seem to be correct, is it?

arnaudsjs (changes requested):
Also add a changelog entry.

server/src/inmantals/server.py
```diff
@@ -54,16 +60,24 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+    async def initialize(self, rootPath, rootUri, capabilities: Dict[str, object], **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
 
         self.rootPath = rootPath
         self.rootUrl = rootUri
         os.chdir(rootPath)
 
+        try:
+            value_set: List[int] = capabilities["workspace"]["symbol"]["symbolKind"]["valueSet"]  # type: ignore
+            self.supported_symbol_kinds = set(map(lsp_types.SymbolKind, value_set))
```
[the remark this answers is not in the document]
sanderr: To your first point: do you mean to load the defaults if none are explicitly passed by the client? I tried that at first but that makes it difficult to know whether it's explicit or implicit. Also, it turned out that I couldn't find a way to make it cleaner that way because I don't see how you could fully dynamic. So, I'm willing to change this but I would need a more concrete suggestion.
- Second point: will fix.

server/src/inmantals/server.py
```diff
@@ -229,6 +251,60 @@ async def textDocument_references(self, textDocument, position, context):  # noq
 
         return [self.convert_location(loc.data) for loc in range]
 
+    async def workspace_symbol(self, query: str) -> List[Dict[str, object]]:
+        if self.types is None:
+            return []
+
+        query_upper: str = query.upper()
```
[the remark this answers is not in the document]
sanderr: I tried to mimic pucharm's behavior (because that's the only reference I've got and it feels useful). Pycharm matches case-insensitively and partially. Would you prefer a different behavior?

server/src/inmantals/server.py
```diff
@@ -229,6 +251,60 @@ async def textDocument_references(self, textDocument, position, context):  # noq
...
+        matching_types: List[Tuple[str, inmanta_type.NamedType]] = [
+            (name, tp)
+            for name, tp in self.types.items()
+            if isinstance(tp, inmanta_type.NamedType) and query_upper in name.upper()
```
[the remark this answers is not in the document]
sanderr: See above

server/tests/test_smoke.py
sanderr: Alas, the location for a plugin is only a `Location`, not a `Range` (in the compiler).

server/tests/test_smoke.py
sanderr: Same here

sanderr: Not exactly the same, but this is i deed the range seg by the compiler.

server/src/inmantals/server.py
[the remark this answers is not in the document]
arnaudsjs: I agree. Matching partially and case-insensitive makes sense.

server/src/inmantals/server.py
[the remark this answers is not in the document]
arnaudsjs: I was not thinking about the scenario where the client doesn't provide any capabilities. I was more thinking about the scenario where the client provides a capability that the extension doesn't support. That would result in an exception right?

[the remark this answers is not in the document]
sanderr: That's right, I hadn't thought of that. I'll change this to ignore (and log a warning) symbol kinds that are not supported by the server.

sanderr:
I'll wait with the changelog entries until the normalization PR has been merged in.

server/tests/test_smoke.py
arnaudsjs: Ok

server/tests/test_smoke.py
arnaudsjs: ok

arnaudsjs (approved):

andraskvr (approved):

PR #35, opened 2020-10-29
Use separate venv for compiler

### LINKED ISSUE #9: use different venv folder, such that ide and cli can use different python version
the compiler creates a venv in the project root

the language server uses the same venv to perform compiles

if they have a different python version this causes trouble

inmanta/inmanta#2466 will add support for a custom venv path. Make sure the language server still works as it did before for older compiler versions.

andraskvr:
closes #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,13 +64,14 @@
         "inmanta.compilerVenv": {
           "type": "string",
           "default": "",
-          "description": "Path to the Python virtual environment used by the compiler. Leave empty to use the default.",
-          "scope": "resource"
+          "description": "Path to the Python virtual environment used by the compiler. Leave empty to use the default. Only available when using Inmanta versions higher than 2020.5 in the language server environment.",
+          "scope": "window"
         }
       }
     }
   },
   "dependencies": {
+    "vsce": "^1.81.1",
     "vscode-languageclient": "^6.1.3"
   },
   "activationEvents": [
--- a/package.json
+++ b/package.json
@@ -60,6 +60,12 @@
           "default": false,
           "description": "Enable the language server",
           "scope": "resource"
+        },
+        "inmanta.compilerVenv": {
+          "type": "string",
+          "default": "",
+          "description": "Path to the Python virtual environment used by the compiler. Leave empty to use the default.",
+          "scope": "resource"
         }
       }
     }
--- a/server/src/inmantals/server.py
+++ b/server/src/inmantals/server.py
@@ -56,6 +56,7 @@ class InmantaLSHandler(JsonRpcHandler):
         self.reverse_anchormap = None
         self.state_lock: asyncio.Lock = asyncio.Lock()
         self.diagnostics_cache: Optional[lsp_types.PublishDiagnosticsParams] = None
+        self.compiler_venv_path: Optional[str] = None
 
     async def initialize(self, rootPath, rootUri, **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
@@ -63,6 +64,9 @@ class InmantaLSHandler(JsonRpcHandler):
         self.rootPath = rootPath
         self.rootUrl = rootUri
         os.chdir(rootPath)
+        init_options = kwargs.get("initializationOptions", None)
+        if init_options:
+            self.compiler_venv_path = init_options.get("compilerVenv", None)
 
         return {
             "capabilities": {
@@ -89,8 +93,10 @@ class InmantaLSHandler(JsonRpcHandler):
             resources.resource.reset()
             handler.Commander.reset()
 
+            if self.compiler_venv_path:
+                logger.debug("Using venv path " + str(self.compiler_venv_path))
             # fresh project
-            Project.set(Project(self.rootPath))
+            Project.set(Project(self.rootPath, venv_path=self.compiler_venv_path))
 
             anchormap = compiler.anchormap()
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -60,6 +60,12 @@
           "default": false,
           "description": "Enable the language server",
           "scope": "resource"
+        },
+        "inmanta.compilerVenv": {
+          "type": "string",
+          "default": "",
+          "description": "Path to the Python virtual environment used by the compiler. Leave empty to use the default.",
+          "scope": "resource"
```
arnaudsjs, line 68: Why scope `resource`? I would expect that `window` matches better. We don't want to change this setting on a per-file or per-directory basis, do we?

server/src/inmantals/server.py
```diff
@@ -89,8 +94,22 @@ def sync_compile_and_anchor() -> None:
...
             # fresh project
-            Project.set(Project(self.rootPath))
+            if (
+                "venv_path" not in project_signature.parameters.keys()
+                and self.compiler_venv_path
+            ):
+                logger.info(
+                    "Compiler virtual env path specified, but "
+                    "the current version of Inmanta doesn't support it, using the default"
+                )
```
arnaudsjs, line 106: This line will produce an incorrect log message. The `extension.ts` makes sure that the value of `compiler_venv_path` is set an a non-empty string. So this log line will report that the report that `compiler_venv_path` is set while it isn't.

arnaudsjs (changes requested):
Also add a changelog entry.

src/extension.ts
```diff
@@ -127,10 +127,18 @@ export function activate(context: ExtensionContext) {
 
 		const errorhandler = new LsErrorHandler(serverOptions)
 
+		let compilerVenv = workspace.getConfiguration('inmanta').compilerVenv;
+		if (!compilerVenv) {
+			compilerVenv = Uri.joinPath(context.globalStorageUri, "compiler-venv").fsPath;
```
sanderr, line 132: Is it ok that this directory is shared among projects? I think it is but I'm not sure. @arnaudsjs what do you think? (I had missed during the core PR review that it was an absolute path.)

server/src/inmantals/server.py
```diff
@@ -56,13 +57,17 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
 
     async def initialize(self, rootPath, rootUri, **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
 
         self.rootPath = rootPath
         self.rootUrl = rootUri
         os.chdir(rootPath)
+        init_options = kwargs.get("initializationOptions", None)
+        if init_options:
+            self.compiler_venv_path = init_options.get("compilerVenv", None)
```
sanderr, line 70: Perhaps default to `os.path.join(self.rootPath, ".env-ls")` or something similar to support other clients than the vscode one.

sanderr (changes requested):

src/extension.ts
arnaudsjs: I don't think it's a good idea to share this venv among projects. Imagine you open two vscode windows using a different inmanta project. If both projects have conflicting requirements defined in the requirements.txt files of their dependencies, both projects will start fighting. Also, I am not convinced that the venv library support concurrent access. It wouldn't surprise me if concurrent access on the same venv would corrupt the venv. 

@sanderr I don't get the remark about the absolute path. The PR on core converts a relative path to an absolute path, as such that a change in the cwd wouldn't cause side effects. What is the link with this comment?

arnaudsjs (changes requested):
See comment

server/src/inmantals/server.py
andraskvr: Good idea, but I'm not sure about the naming. 
`.env-ls` kind of suggests a bit that the ls is installed there
What I wrote now `.env-ls-compiler`, doesn't really make it clear either.
Any suggestions @sanderr ?

src/extension.ts
sanderr: Now that I think about it the remark about the absolute path does not have anything to do with this. I noticed it while typing the comment and it confused my thinking I guess. And indeed, it's not an absolute path, it's relative to the current working directory, while I had assumed relative to the project root.
Anyway, you can safely ignore that part. I agree with @arnaudsjs' assessment.

server/src/inmantals/server.py
sanderr: You're right. I think `.env-ls-compiler` is sufficiently clear, though it's not very concise but I guess that doesn't really matter. On the other hand, I think 99% of people working with inmanta will either use the language server with vscode or not at all, so the naming for this isn't terribly important. I'm fine with either one, no others come to mind.

andraskvr:
Changelog still to be added after #36

server/src/inmantals/server.py
```diff
@@ -67,7 +67,7 @@ async def initialize(self, rootPath, rootUri, **kwargs):  # noqa: N803
         os.chdir(rootPath)
         init_options = kwargs.get("initializationOptions", None)
         if init_options:
-            self.compiler_venv_path = init_options.get("compilerVenv", None)
+            self.compiler_venv_path = init_options.get("compilerVenv", ".env-ls-compiler")
```
sanderr, line 70: The venv path is relative to the current working directory, not to the project root: `os.path.join(self.rootPath, ".env-ls-compiler")`

sanderr (changes requested):

arnaudsjs (approved):
Agree with Sander's comment

sanderr (approved):

PR #36, opened 2020-10-29
normalized changelog format and fixed incorrect versions

sanderr:
### Changes to server changelog
- changelog now conforms to the inmanta-tools format
- current version added
- previous version was listed as 0.3.3 but `setup.py` was at 0.2.0 at the time of those changes.

### Changes to client changelog
- changelog now conforms to the inmanta-tools format
- current version added

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,12 +1,10 @@
-# Change Log
+# v 0.2.1 (?)
+Changes in this release:
+- Added language server to client diagnostics notifications to highlight errors (#18)
 
-[Changelog](https://github.com/inmanta/vscode-inmanta/server/CHANGELOG.md)
+# v 0.2.0
 
-Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
-
-## [0.3.3]
-
-### Engineering
+## Engineering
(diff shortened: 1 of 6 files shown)


server/CHANGELOG.md
```diff
@@ -1,12 +1,10 @@
-# Change Log
+# v 0.2.1 (?)
```
andraskvr, line 1: 0.2.1 is already released:
https://pypi.org/project/inmantals/

andraskvr (changes requested):

server/CHANGELOG.md
sanderr: I see. Does this mean I should update `setup.py` to contain 0.2.2?

arnaudsjs: Our other release pipelines bump the version on master to the next minor release. To be consistent, it should be 0.3.0.

arnaudsjs (changes requested):
See comment

server/CHANGELOG.md
sanderr: Right, we added new features anyway, my reasoning was a bit off.

andraskvr (approved):

arnaudsjs (approved):

PR #47, opened 2020-11-03
Extend language server support on windows

### LINKED ISSUE #26: Support language server on windows
Currently VSCode does not work on windows. The issue seems to be that the pipe language server is not supported. 

It is probably best to use socket on all platforms?

If you do not have a windows to test it, ask for credentials on windows.ii.inmanta.com (from Bart)

andraskvr:
Closes #26


server/src/inmantals/tcpserver.py
```diff
@@ -35,7 +39,12 @@ def main():
     logging.basicConfig(level=logging.DEBUG)
 
     server = JsonRpcServer(InmantaLSHandler)
-    server.listen(5432)
+    port = int(sys.argv[1]) if len(sys.argv) > 1 else 5432
+    sys.stdout.write(f"starting server on port {port}")
+    sys.stdout.flush()
+    sys.stderr.write(f"Log file can be found at {logfile}")
+    sys.stderr.flush()
```
sanderr, line 46: I'm not sure we should log this to stderr, stdout seems like the conventional stream for a message like this. Any specific reason to log to stderr?

sanderr (changes requested):

server/src/inmantals/tcpserver.py
andraskvr: Currently, only the stderr is displayed to the user in the output channel in vscode (with both the pipe and tcp server), that's why I've logged it there. But I agree that it would be better in stdout. Do you think we should also show the stdout in the output channel?

sanderr: I'm not sure. I think it would be ok to show stdout as well. @arnaudsjs what do you think?

arnaudsjs: I don't have a strong opinion on this, but I agree that stdout is more conventional. 

What is more strange to me is the fact that the first log entry is logged to stdout and the second to stderr?

src/extension.ts
```diff
@@ -5,62 +5,117 @@ import * as net from 'net';
...
+			initializationOptions: {
+				compilerVenv: compilerVenv
+			}
 		};
+		return clientOptions;
+	}
+	
+	async function startTcp(clientOptions: LanguageClientOptions) {
+		const pp: string = createVenvIfNotExists();
+		const serverPort = await getPort();
```
arnaudsjs, line 46: On which address does this return free port? IPv4/IPv6? On which address?

src/extension.ts
```diff
@@ -5,62 +5,117 @@ import * as net from 'net';
...
+		
+		const serverProcess = cp.spawn(pp, ["-m", "inmantals.tcpserver", serverPort.toString()]);
+		let serverStdout = '';
+		serverProcess.stdout.on('data', (data) => {
+			lsOutputChannel.appendLine(`stdout: ${data}`);
+			serverStdout += data;
```
arnaudsjs, line 52: The size of this string only grows. This is a memory leak.

server/src/inmantals/tcpserver.py
andraskvr: Right, I changed it so both use stdout

src/extension.ts
```diff
@@ -168,10 +218,10 @@ export function activate(context: ExtensionContext) {
 			window.showInformationMessage(`No Python3 interpreter found at "${pp}". Falling back to default virtual environment`);
 		}
 		const venvBaseDir = Uri.joinPath(context.globalStorageUri, ".env").fsPath;
-		const venvPath = Uri.joinPath(context.globalStorageUri, ".env", "bin", "python3").fsPath;
+		const venvPath = getDefaultVenvPath();
 
 		if (!fs.existsSync(venvBaseDir)) {
-			const venvProcess = cp.spawnSync("python3", ["-m", "venv", venvBaseDir]);
+			const venvProcess = cp.spawnSync("python", ["-m", "venv", venvBaseDir]);
```
arnaudsjs, line 224: I don't think this is a good idea. Is it require to make this work on Windows?

sanderr (approved):

arnaudsjs (changes requested):

src/extension.ts
andraskvr: At first I thought so, but it turned out that I just needed to restart, nice catch.

package.json
```diff
@@ -115,6 +116,7 @@
     "fs-extra": "^9.0.1",
     "glob": "^7.1.6",
     "mocha": "^8.1.3",
+    "rimraf": "^3.0.2",
```
wouterdb, line 119: what is this?

wouterdb (approved):

package.json
andraskvr: Basically `rm -rf` in node js.
This way it works on windows as well.

src/extension.ts
```diff
@@ -5,62 +5,122 @@ import * as net from 'net';
...
+			revealOutputChannelOn: RevealOutputChannelOn.Info,
+			initializationOptions: {
+				compilerVenv: compilerVenv
+			}
 		};
+		return clientOptions;
+	}
+	
+	async function startTcp(clientOptions: LanguageClientOptions) {
+		const host = "localhost";
```
arnaudsjs, line 45: Wouldn't it be better to use `127.0.0.1` instead of `localhost`. In the past we had issues on core, because localhost must be translated to an address. This might be an IPv4 or an IPv6 address, but you never know which. In this case three independent libraries will do that translation:
* the `getPort()` call,
* the `tcpServer()` and
* the `net.connect()`

If one of those libraries translate localhost in a different way, there is an issue. @wouterdb What do you think?

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #48, opened 2020-11-03
Issue/39 venv installation test

### LINKED ISSUE #39: Add venv installation test to the typescript part of the extension
Issue splitted from https://github.com/inmanta/vscode-inmanta/issues/17

edvgui:
Changes in this MR: 
1. Added tests for compiler virtual environment creation.  
2. Testing automatic creation of extension venv and installation of LS, by using it while running the tests.

Closes #39


src/test/suite/compile.test.ts
```diff
@@ -57,41 +38,46 @@ describe('Compile checks', () => {
...
+			const succeeded = await waitForCompile(logPath, 20000);
+			assert.strictEqual(succeeded, test.succeed, `The model should ${test.succeed ? "" : "not"} compile, but did ${succeeded ? "" : "not"}.`);
+
+			const libsExists = fs.pathExistsSync(libsPath);
+			assert.strictEqual(libsExists, true, "The libs folder hasn't been created");
+
+			const pythonPath: string = workspace.getConfiguration('inmanta').get<string>('pythonPath');
+			const compilerVenv: string = workspace.getConfiguration('inmanta').get<string>('compilerVenv');
+			const inmantaVersion: string = await getInmantaVersion(pythonPath);
+			envPath = compareVersions(inmantaVersion, "2020.5") <= 0 ? path.resolve(workspaceUri.fsPath, '.env') : compilerVenv;
```
sanderr, line 69: Perhaps use [`semver.coerce`](https://www.npmjs.com/package/semver) instead? It doesn't seem to handle dev versions very well but that can be easily worked around if you just compare major and minor.

sanderr (changes requested):

src/test/suite/compile.test.ts
edvgui: Not sure why we should use it if it doesn't handle the dev versions? The custom method does.

sanderr: I don't think it's fully stable yet. I uses `splitedX.length` to determine which version comes first but `2020.6.dev0 < 2020.6` while `2020.6.1 > 2020.6`. The part after that has similar issues. These issues could be fixed, but I think it would be easier to just use the library, which is already in `package-lock.json` anyway.
In any case, I don't feel very strongly about the above, but I would definitely limit the version comparison to major and minor, because if we would do any 2020.5.x patch release, these should be handled the same as 2020.5. All 2020.6.x or 2020.6.devx should be handled the same as well, so I think it's sufficient (better even) to just check major and minor against 2020.5. You'll probably have to rename the function a bit to convey this idea.

edvgui: Ow, interesting, I didn't knew that `2020.6.dev0 < 2020.6`.  I will switch to the library then, and limit to major and minor as you suggested.

edvgui: I wanted to test `coerce` a bit:
```typescript
import { coerce } from 'semver';

const versions = [
    "2020.5",
    "2020.5.1",
    "2020.5.dev0",
    "2020.6",
    "2020.6.dev0",
    "2019.1"
]

function cleanVersion(version: string): string {
    const splited = version.split(".");
    const major = splited.length > 0 ? splited[0] : "0";
    const minor = splited.length > 1 ? splited[1] : "0";
    return [major, minor].join(".");
}

versions.forEach(versionA => versions.forEach(versionB => {
    console.log(`${versionA} <= ${versionB} ? ${coerce(versionA) <= coerce(versionB)}`);

    const cleanedVersionA = cleanVersion(versionA);
    const cleanedVersionB = cleanVersion(versionB);
    console.log(`${cleanedVersionA} <= ${cleanedVersionB} ? ${coerce(cleanedVersionA) <= coerce(cleanedVersionB)}`);
}));
```
<details>

<summary>It gives the output:</summary>

```
2020.5 <= 2020.5 ? true
2020.5 <= 2020.5 ? true
2020.5 <= 2020.5.1 ? true
2020.5 <= 2020.5 ? true
2020.5 <= 2020.5.dev0 ? true
2020.5 <= 2020.5 ? true
2020.5 <= 2020.6 ? true
2020.5 <= 2020.6 ? true
2020.5 <= 2020.6.dev0 ? true
2020.5 <= 2020.6 ? true
2020.5 <= 2019.1 ? false
2020.5 <= 2019.1 ? false
2020.5.1 <= 2020.5 ? false
2020.5 <= 2020.5 ? true
2020.5.1 <= 2020.5.1 ? true
2020.5 <= 2020.5 ? true
2020.5.1 <= 2020.5.dev0 ? false
2020.5 <= 2020.5 ? true
2020.5.1 <= 2020.6 ? true
2020.5 <= 2020.6 ? true
2020.5.1 <= 2020.6.dev0 ? true
2020.5 <= 2020.6 ? true
2020.5.1 <= 2019.1 ? false
2020.5 <= 2019.1 ? false
2020.5.dev0 <= 2020.5 ? true
2020.5 <= 2020.5 ? true
2020.5.dev0 <= 2020.5.1 ? true
2020.5 <= 2020.5 ? true
2020.5.dev0 <= 2020.5.dev0 ? true
2020.5 <= 2020.5 ? true
2020.5.dev0 <= 2020.6 ? true
2020.5 <= 2020.6 ? true
2020.5.dev0 <= 2020.6.dev0 ? true
2020.5 <= 2020.6 ? true
2020.5.dev0 <= 2019.1 ? false
2020.5 <= 2019.1 ? false
2020.6 <= 2020.5 ? false
2020.6 <= 2020.5 ? false
2020.6 <= 2020.5.1 ? false
2020.6 <= 2020.5 ? false
2020.6 <= 2020.5.dev0 ? false
2020.6 <= 2020.5 ? false
2020.6 <= 2020.6 ? true
2020.6 <= 2020.6 ? true
2020.6 <= 2020.6.dev0 ? true
2020.6 <= 2020.6 ? true
2020.6 <= 2019.1 ? false
2020.6 <= 2019.1 ? false
2020.6.dev0 <= 2020.5 ? false
2020.6 <= 2020.5 ? false
2020.6.dev0 <= 2020.5.1 ? false
2020.6 <= 2020.5 ? false
2020.6.dev0 <= 2020.5.dev0 ? false
2020.6 <= 2020.5 ? false
2020.6.dev0 <= 2020.6 ? true
2020.6 <= 2020.6 ? true
2020.6.dev0 <= 2020.6.dev0 ? true
2020.6 <= 2020.6 ? true
2020.6.dev0 <= 2019.1 ? false
2020.6 <= 2019.1 ? false
2019.1 <= 2020.5 ? true
2019.1 <= 2020.5 ? true
2019.1 <= 2020.5.1 ? true
2019.1 <= 2020.5 ? true
2019.1 <= 2020.5.dev0 ? true
2019.1 <= 2020.5 ? true
2019.1 <= 2020.6 ? true
2019.1 <= 2020.6 ? true
2019.1 <= 2020.6.dev0 ? true
2019.1 <= 2020.6 ? true
2019.1 <= 2019.1 ? true
2019.1 <= 2019.1 ? true
```
</details>

We agree that the cleaned version gives the result we want?

sanderr: Indeed, though I think it would be easier still to just do
```
version_obj = semver.coerce(version)
version_obj.major > 2020 or version_obj.major == 2020 and version_obj.minor > 5
```
or the other way around depending on what you need.

sanderr: (I shouldn't snake_case, this is ts, not Python)

sanderr (approved):

src/test/suite/compile.test.ts
edvgui: Did it in https://github.com/inmanta/vscode-inmanta/pull/48/commits/e7864770dfc7437932d10f540620995442248afa

andraskvr (approved):

PR #54, opened 2020-11-04
Issue/32 open file

### LINKED ISSUE #32: Opening a model file without opening the project folder causes unintuitive error
According to the spec this passes `rootPath=None` to the `initialize` call. This causes the server to raise the following exception:
"chdir: path should be string, bytes, os.PathLike or integer, not NoneType". The `initialize` method should be updated to check for `rootPath=None` and raise an exception that makes clear to the user that an inmanta configuration file only has meaning within the context of an inmanta project.

arnaudsjs:
Closes #32


src/extension.ts
```diff
@@ -16,7 +16,12 @@ export async function activate(context: ExtensionContext) {
 	let lsOutputChannel = null;
 
 	async function startServerAndClient() {
-		const clientOptions = await getClientOptions();
+		var clientOptions;
```
andraskvr, line 19:
```suggestion
		let clientOptions;
```

src/extension.ts
```diff
@@ -25,10 +30,18 @@ export async function activate(context: ExtensionContext) {
 	}
 
 	async function getClientOptions(): Promise<LanguageClientOptions> {
+		var compilerVenv: string = workspace.getConfiguration('inmanta').compilerVenv;
+		if (!compilerVenv) {
+			if(context.storageUri == undefined){
+				window.showWarningMessage("A folder should be opened instead of a file in order to use the inmanta extension.");
+				throw "A folder should be opened instead of a file in order to use the inmanta extension.";
```
andraskvr, line 37:
```suggestion
				throw Error("A folder should be opened instead of a file in order to use the inmanta extension.");
```

src/extension.ts
```diff
@@ -25,10 +30,18 @@ export async function activate(context: ExtensionContext) {
 	}
 
 	async function getClientOptions(): Promise<LanguageClientOptions> {
+		var compilerVenv: string = workspace.getConfiguration('inmanta').compilerVenv;
```
andraskvr, line 33:
```suggestion
		let compilerVenv: string = workspace.getConfiguration('inmanta').compilerVenv;
```

src/test/suite/loadExtension.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as path from 'path';
+
+import { Uri, window, commands, workspace, TextDocument, TextEditor, extensions } from 'vscode';
+
+const cfFile: Uri = Uri.file(path.resolve(__dirname, '../../../src/test/workspace/valid.cf'));
+
+describe('Load extension', () => {
+
+    beforeEach(async function() {
+        // await commands.executeCommand('workbench.action.closeWorkspace');
```
andraskvr, line 12: Comment leftover

andraskvr (changes requested):

andraskvr (approved):

server/src/inmantals/pipeserver.py
```diff
@@ -38,12 +38,12 @@ def main():
     stdin = PipeIOStream(sys.stdin.fileno())
     stdout = PipeIOStream(sys.stdout.fileno())
     handler = InmantaLSHandler(stdin, stdout, "0.0.0.0")
-    sys.stderr.write("starting")
+    sys.stderr.write("starting\n")
```
sanderr, line 41: Write these to stdout instead? The client now logs stdout as well.

sanderr (changes requested):

server/src/inmantals/pipeserver.py
andraskvr: I think that's only true for the `tcpserver`

sanderr: Right, nevermind then.

sanderr (approved):

PR #77, opened 2021-02-10
fixed release pipeline issue related to npm install artifacts

sanderr:
Installing vsce would update `package.json` and `package-lock.json`. This would cause issues with the checkout commands following it. Simply forcing checkout (cleaning up the changes) would cause problems when running `vsce`.

### DIFF
--- a/.ci/Jenkinsfile-extension-vsce
+++ b/.ci/Jenkinsfile-extension-vsce
@@ -44,6 +44,9 @@ pipeline {
                             echo 'echo \${GITHUB_TOKEN}' > ${GIT_ASKPASS}
                             chmod u+x ${GIT_ASKPASS}
 
+                            # install the extension for its vsce dependency, required for building and publishing the extension
+                            npm ci
+
                             # checkout appropriate branch
                             if [ "${version_increment}" == "none" ] || [ "${version_increment}" == "patch" ]; then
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile-extension-vsce
```diff
@@ -162,6 +159,17 @@ pipeline {
             }
         }
 
+        stage("Install vsce") {
+            when {
+                expression { "${run_job}" == "true" }
+            }
+            steps {
+                dir('vscode-inmanta') {
+                    sh 'npm install vsce'
```
andraskvr, line 168: Since `vsce` is a dependency of the `vscode-inmanta` extension, I think it would be better to install the extension itself. When you use the `npm ci` command, it will 
1. clean the current node_modules 
2. install everything according to the `package-lock.json`, without changing anything, so the cleanup step shouldn't be necessary.

andraskvr (changes requested):

.ci/Jenkinsfile-extension-vsce
sanderr: That's interesting, I'll do that, thanks.

andraskvr (approved):

arnaudsjs (approved):

PR #83, opened 2021-05-28
Migrate to inmanta-core dependency

arnaudsjs:
Replace the dependency on inmanta with a dependency on inmanta-core


sanderr (approved):
Nice that you updated the code as well, I hadn't thought of that.

wouterdb (approved):

PR #266, opened 2021-09-07
Fix hanging test loadExtension

arnaudsjs:
This PR fixes the following issues:

* Make sure that the `LoadExtensions` test doesn't hang anymore
* Fix some styling issues
* Remove redundant code that configures the `settings.json` file on a test project
* Add the `--disable-gpu` flag to reduce the number of CPU-related error/warning in the log.

### DIFF
--- a/src/test/loadExtension/loadExtension.test.ts
+++ b/src/test/loadExtension/loadExtension.test.ts
@@ -9,16 +9,16 @@ const cfFile: Uri = Uri.file(path.resolve(__dirname, '../../../src/test/compile/
 describe('Load extension', () => {
 
     beforeEach(async function() {
-        await commands.executeCommand('workbench.action.closeFolder');
+        await commands.executeCommand('workbench.action.closeActiveEditor');
     });
 
-    it('Load .cf file wihtout opening a directory NO compilerVenv configured', async function() {
+    it('Load .cf file without opening a directory NO compilerVenv configured', async function() {
         await workspace.getConfiguration('inmanta').update('compilerVenv', "", true);
         // Verify initial state
         assert.ok(!extensions.getExtension('inmanta.inmanta').isActive);
 
         // Open a single file instead of a folder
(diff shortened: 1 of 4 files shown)


src/test/loadExtension/loadExtension.test.ts
```diff
@@ -9,16 +9,16 @@ const cfFile: Uri = Uri.file(path.resolve(__dirname, '../../../src/test/compile/
 describe('Load extension', () => {
 
     beforeEach(async function() {
-        await commands.executeCommand('workbench.action.closeFolder');
+        await commands.executeCommand('workbench.action.closeActiveEditor');
```
arnaudsjs, line 12: This line made the test case hang. Actually I cannot come up with a reason why this line should be there, but it also exists in the other test cases. When I remove this line, the test cases succeed. @sanderr @andraskvr Any idea why this line should be there.

sanderr: I don't know eiter.

sanderr (commented):
Could you elaborate on what makes the writing of the settings redundant?

src/test/loadExtension/loadExtension.test.ts
andraskvr: I think it might have been put there to make sure the tests start from a clean slate without an open editor, even if previous ones fail - but I'm not sure. If it's not necessary (anymore, maybe because of changes to vscode itself), we can remove it

sanderr (approved):

arnaudsjs:
> Could you elaborate on what makes the writing of the settings redundant?

* The default value of `inmanta.ls.enabled` is True
* The other settings `INMANTA_PYTHON_PATH` and `INMANTA_COMPILER_VENV` don't exist anymore.

andraskvr (approved):

sanderr (approved):

PR #378, opened 2021-12-16
Updated language server to use encapsulating venv for core>=6

sanderr:
This is the server part for inmanta/inmanta-core#3080 (merging into the #374 branch).

Should I add unit tests? I think it wouldn't be straightforward to test this on its own and we already run the test suite against both iso4 and iso5 so we've got end-to-end tests of a sort already.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -1,6 +1,10 @@
 # Inmanta Language server
 
-provides the bridge between the Inmanta compiler and Visual Studio Code ide, by implementing a Language Server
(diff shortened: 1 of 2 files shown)


wouterdb (approved):

wouterdb:
I wouldn't add unit tests if this is covered by existing tests

FloLey (approved):

arnaudsjs (approved):

sanderr:
Merged manually in 915a58c and 17b1252.

PR #404, opened 2022-01-13
fix issue @types/vscode version

FloLey:
This commit fixes the following issue:

ERROR  @types/vscode ^1.63.0 greater than engines.vscode ^1.50.0. Consider upgrade engines.vscode or use an older @types/vscode version

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,5 @@ updates:
   ignore:
     - dependency-name: "get-port"
       update-types: ["version-update:semver-major"]
+    - dependency-name: "@types/vscode"
+      update-types: ["version-update:semver-minor"]


sanderr (changes requested):
You'll need to update the dependabot config at `.github/dependabot.yml` to configure it to ignore this version. You can look at the `get-port` constraint for an example (we'll have to be a bit more strict and ignore minors as well for this one). Then could you add it to inmanta/infra-tickets#121 afterwards?

Stijnkool:
Why not upgrade the engines version? Downgrading the type definitions is also fine and very safe.
Upgrading the engines might not be safe. But im not really in the loop with this stuff 🤷

Stijnkool (approved):

FloLey:
@Stijnkool I just decided to do what seemed safest to me. And changing a devDependency should have little consequences.

sanderr (approved):

Issue #442, opened 2022-02-02
Missing dependencies when Installing the extension 


FloLey:
After installing the vscode extension I had to install by hand the following 3 packages and there own missing dependencies:

- lru-cache(6.0.0):
   - yallist(4.0.0)
- log-symbols(4.1.0)
- minimatch(3.0.4):
   - brace-expansion(1.1.17):
      - balanced-match (1.0.2)
       - concat-map (0.0.1)

once done it was working again.

PR #443, opened 2022-02-02
fix missing dependencies

FloLey:
When install the extension in vscode the activation fails due to some missing dependencies and dependencies of those dependencies.
- lru-cache(6.0.0):
   - yallist(4.0.0)
- log-symbols(4.1.0)
- minimatch(3.0.4):
   - brace-expansion(1.1.17):
      - balanced-match (1.0.2)
       - concat-map (0.0.1)

This commit fixes the issue by adding the missing dependencies. 

issue #442

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
             "license": "Apache-2.0",
             "dependencies": {
                 "get-port": "^5.1.1",
+                "log-symbols": "^4.1.0",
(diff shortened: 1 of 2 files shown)


Stijnkool (approved):
I have no real context here. But updating dependencies is fine.

arnaudsjs (approved):

PR #473, opened 2022-02-14
test inmantals against iso5-stable

sanderr:


server/tox.ini
```diff
@@ -18,6 +18,7 @@ deps=
     inmlatest: inmanta-core
     inmmaster: git+https://github.com/inmanta/inmanta-core.git
     inmiso4-stable: git+https://github.com/inmanta/inmanta-core.git@iso4-stable
+    inmiso5-stable: git+https://github.com/inmanta/inmanta-core.git@iso5-stable
```
arnaudsjs, line 21: * We need to add `iso5-stable` to `tox.envlist` as well.
* `tox.envlist` indicates that the tests run using `python3.6` and `python3.8`. I don't think that's accurate.

arnaudsjs (changes requested):

server/tox.ini
sanderr: Think I fixed both.

server/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = pep8,py{36,38}-inm{master,iso4-stable,latest}
+envlist = pep8,py36-inmiso4,py39-inm{master,iso{4,5}-stable,latest}
```
arnaudsjs, line 2: Why also run `iso4` using `python39`?

arnaudsjs (changes requested):

server/tox.ini
sanderr: Good point, it's not OSS so I guess 3.6 suffices.

arnaudsjs (approved):

PR #620, opened 2022-05-25
Fix potential race condition in restart behavior

### LINKED ISSUE #594: Race condition in restart behavior of language server and client
The `extension.ts` has logic to restart the language server and client when the inmanta or pythonExtension configuration changes. This restart logic is asynchronous. As such, a locking mechanism is required to make sure that two restart requests don't interfere with eachother. This locking mechanism is currently absent.

I did some small tests and managed to trigger the issue in the following way:

* Make sure the inmanta language server is running
* Change to another Python environment
* Quickly change once more to another Python environment

Result: Several pipeservers are running side-by-side while only one should be running

```
 $ ps -auxf|grep -i pipeserver
arnaud     43530 10.2  0.2 376052 68512 ?        Sl   08:18   0:00  |           \_ /home/arnaud/.virtualenvs/project/bin/python -m inmantals.pipeserver
arnaud     43551 20.5  0.2 376052 68620 ?        Sl   08:18   0:00  |           \_ /home/arnaud/.virtualenvs/test/bin/python -m inmantals.pipeserver
```

andraskvr:
closes #594


FloLey (approved):
nice:)

PR #846, opened 2022-11-23
Fix: remove EntityLike

FloLey:
remove EntityLike as it doesn't exist anymore.


FloLey:
I changed a test but I'm not sure this is the behavior we expect.

sanderr (changes requested):

sanderr (approved):

PR #879, opened 2022-12-16
Fix broken tests

arnaudsjs:
The log of the tests run showed that every time a language server was started, it was immediately restart. While the configured venv didn't change. Combined with the bug where two language servers would run in parallel for a short period of time, this caused the tests to fail.

### DIFF
--- a/src/python_extension.ts
+++ b/src/python_extension.ts
@@ -24,9 +24,9 @@ export class PythonExtension {
 				if(this.executionDetails.execCommand[0] !== newExecutionDetails.execCommand[0]){
 					log(`Active interpreter changed for: ${resource}`);
 					log(`Execution details: ${JSON.stringify(this.executionDetails)}`);
+					this.executionDetails = newExecutionDetails;
 					this.onChangeCallback();
 				}
-				this.executionDetails = newExecutionDetails;
 			}
 		);
 	}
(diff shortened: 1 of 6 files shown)


src/python_extension.ts
```diff
@@ -20,14 +20,16 @@ export class PythonExtension {
...
-				log(`Execution details: ${JSON.stringify(this.executionDetails)}`);
-				this.executionDetails = pythonApi.settings.getExecutionDetails(resource);
-				this.onChangeCallback();
+				let newExecutionDetails = pythonApi.settings.getExecutionDetails(resource);
+				if(this.executionDetails.execCommand[0] !== newExecutionDetails.execCommand[0]){
+					log(`Active interpreter changed for: ${resource}`);
+					log(`Execution details: ${JSON.stringify(this.executionDetails)}`);
+					this.onChangeCallback();
+				}
+				this.executionDetails = newExecutionDetails;
```
sanderr, line 29: The respective order of these two statements changed. I don't know what `onChangeCallback` does so I can't really estimate if this is a potential issue.

sanderr (approved):

src/python_extension.ts
arnaudsjs: I don't think it matters, but it's easy to use the same order so I made the change.

Hugo-Inmanta (approved):

PR #939, opened 2023-01-27
Issue/545 python extension bottom bar

FloLey:
Add information about how to show the python interpreter information in status bar


FloLey:
I think that's all there is to do?

sanderr (approved):

README.md
```diff
@@ -68,6 +68,8 @@ The extension currently has two settings:
 The plugin uses the Python extension to change environments. This can be done by clicking on the active environment in the bottombar
 and then selecting the environment to activate.
 
+If the environment is not shown in the bottom bar, change the setting "python.interpreter.infoVisibility" to "Always".
```
wouterdb, line 71: Can't we just set this? or is that considered impolite?

FloLey: I don't think I'd like it if a plugin would alter my settings. I would get frustrated If I change the setting and for some reason my changes are always overwritten and I wouldn't know why this is happening.
If we still decide to do this, I propose to change the setting on activation, and set it back to the previous value on deactivation. 
I haven't verified (yet) if it is doable but I suppose it is.

FloLey: @wouterdb Would you want me to try?

FloLey:
@wouterdb I think this changes makes the environment show in the bottombar when a .cf file is opened. I test it manually and it seemed to work.

wouterdb (changes requested):
Nice!

When trying this, I have the following things

1. I think this message has to be updated. It would be usefull if it could offer a button that activates the `python: select interperter` command (that you can activate via ctrl-shift-p
image (command id: python.setInterpreter)

(My dev setup might be a bit crapy, so you may have to verify if this dialog also happens to you)

Otherwise: very nice.

FloLey:
@wouterdb I don't find how to get your dialog but I find it really strang:
inmanta.pythonPath was removed here: https://github.com/inmanta/vscode-inmanta/commit/642c1562fe2b9704834360def91c6e705d5e6e37#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L48
and I don't find any reference to it in the code.

when I have no interpreter selected I get this:

image

I'll try to change this to have a button to activate an interpreter and I hope i'll get luck and it will fix your issue at the same time

FloLey:
@wouterdb I fixed my dialog, could you try yours again?

wouterdb (approved):

wouterdb:
could not reproduce

Issue #891, opened 2022-12-21
Allow working on modules


wouterdb:
- When a module is opened in vscode, detect this and create a project somewhere to be able to load all files of the module.
(cfr pytest-inmanta)
- only look at root dir
- Add options to set the repos to pull from in the configuration

bartv:
This will not work for modules like connect because you might not have access to all dependencies. An option to handle this, is using something like python entrypoints to indicate which modules you want to load.

wouterdb:
Good remark, we could load the _init and all open files and provide a config in vscode to override?

bartv:
I think it is also import to still make things work when one import does not exist. For example vscode keeps on working even if an import does not exist. This is something we should do in the language server as well

[Screencast from 2022-12-21 17-08-50.webm](https://user-images.githubusercontent.com/10819/208950464-b06c786c-5f51-4824-aedc-cb0a5a8792b9.webm)

PR #992, opened 2023-03-01
Issue/891 allow working on modules

Hugo-Inmanta:
#closes https://github.com/inmanta/vscode-inmanta/issues/891

This PR adds support for working on a module. (Previously only projects were supported)

- Behind the scenes it creates a project in the /tmp/vs-code with a main.cf that only contains `import <module_name>`.
- The `repo` setting of the project.yml file of this project can be configured via the newly added extension setting `repos`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
                 "inmanta.repos": {
                     "type": "string",
                     "default": "",
-                    "markdownDescription": "Optional yaml list of repositories to indicate the inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, the inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml) file.",
+                    "markdownDescription": "Optional yaml list of repositories to indicate the Inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml) file.",
                     "scope": "window"
                 }
             }
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
         "get-port": "^5.1.1",
         "log-symbols": "^5.1.0",
         "lru-cache": "^7.17.0",
-        "minimatch": "^5.1.2 || ^7.1.1",
+        "minimatch": "^5.1.2 || ^7.3.0",
         "vscode-languageclient": "^8.1.0"
     },
     "activationEvents": [
@@ -121,8 +121,8 @@
         "@types/node": "^18.14.2",
         "@types/semver": "^7.3.13",
         "@types/vscode": "^1.63.2",
-        "@typescript-eslint/eslint-plugin": "^5.53.0",
-        "@typescript-eslint/parser": "^5.53.0",
+        "@typescript-eslint/eslint-plugin": "^5.54.0",
+        "@typescript-eslint/parser": "^5.54.0",
         "@vscode/test-electron": "^2.2.3",
         "@vscode/vsce": "^2.18.0",
         "eslint": "^8.35.0",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,12 @@
                     "default": "",
                     "description": "Path to the Python virtual environment used by the compiler. Will only be used by language server if inmanta-core<6. Leave empty to use the default. Only available when using Inmanta versions higher than 2020.5 in the language server environment.",
                     "scope": "window"
+                },
+                "inmanta.repos": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "Optional yaml list of repositories to indicate the inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, the inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml) file.",
+                    "scope": "window"
                 }
             }
         },
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
         "@typescript-eslint/parser": "^5.53.0",
         "@vscode/test-electron": "^2.2.3",
         "@vscode/vsce": "^2.18.0",
-        "eslint": "^8.34.0",
+        "eslint": "^8.35.0",
         "eslint-config-google": "^0.14.0",
         "fs-extra": "^11.1.0",
         "glob": "^8.1.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -60,6 +60,12 @@
                     "default": "",
                     "description": "Path to the Python virtual environment used by the compiler. Will only be used by language server if inmanta-core<6. Leave empty to use the default. Only available when using Inmanta versions higher than 2020.5 in the language server environment.",
                     "scope": "window"
+                },
+                "inmanta.repos": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "Optional yaml list of repositories to indicate the inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, the inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml) file.",
```
Hugo-Inmanta, line 67:
```suggestion
                    "markdownDescription": "Optional yaml list of repositories to indicate the Inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml) file.",
```

server/src/inmantals/pipeserver.py
```diff
@@ -37,7 +37,7 @@ def main():
     try:
         IOLoop.current().run_sync(handler.start)
     except Exception as e:
-        logging.error(e)
+        logger.error(e)
```
Hugo-Inmanta, line 40: This is correct right ?

server/src/inmantals/server.py
```diff
@@ -118,22 +125,83 @@ def flatten(self, line, char):
         assert char < 100000
         return line * 100000 + char
 
+    def create_tmp_project(self) -> str:
+        ten_random_digits = "".join(random.choice(string.digits) for _ in range(10))
+        project_name = "project" + ten_random_digits
```
Hugo-Inmanta, line 130: I'm not 100% sure about this random name thing. One the one hand it allows opening 2 different modules in 2 different windows. But on the other hand I think my cleaning logic is not good enough at the moment and we are leaking some projects that are never cleaned up in case the `shutdown` method isn't called.

server/src/inmantals/server.py
```diff
@@ -118,22 +125,83 @@ def flatten(self, line, char):
...
+        modulepath = ["libs", os.path.dirname(self.rootPath)]
+
+        with open(os.path.join(self.project_dir, "project.yml"), "w+") as fd:
+            metadata: typing.Mapping[str, object] = {
+                "name": "testcase",
+                "description": "Project for testcase",
+                "repo": yaml.safe_load(self.repos),
+                "modulepath": modulepath,
+                "downloadpath": "libs",
+                "install_mode": install_mode.value,
```
Hugo-Inmanta, line 149: Not too sure about these two

Hugo-Inmanta:
As far as testing goes, I tried opening a simple module (tested with both v1 and v2) and it looks like ctrl-click navigation works

server/src/inmantals/pipeserver.py
wouterdb: yes

server/src/inmantals/server.py
```diff
@@ -118,22 +125,83 @@ def flatten(self, line, char):
         assert char < 100000
         return line * 100000 + char
 
+    def create_tmp_project(self) -> str:
+        ten_random_digits = "".join(random.choice(string.digits) for _ in range(10))
+        project_name = "project" + ten_random_digits
+        self.project_dir = os.path.join("/tmp", "vs-code", project_name)
```
wouterdb, line 131: I think we may have to use the tmp folder of vscode (i.e. on windows, ...) 

I don't know if vscode has any guidelines? 

Otherwise I would use the tempfile of python, which should work on all platforms 
https://docs.python.org/3/library/tempfile.html

server/src/inmantals/server.py
```diff
@@ -118,22 +125,83 @@ def flatten(self, line, char):
...
+
+        install_mode = InstallMode.master
+
+        modulepath = ["libs", os.path.dirname(self.rootPath)]
+
+        with open(os.path.join(self.project_dir, "project.yml"), "w+") as fd:
+            metadata: typing.Mapping[str, object] = {
+                "name": "testcase",
+                "description": "Project for testcase",
+                "repo": yaml.safe_load(self.repos),
```
wouterdb, line 146: how do you know if it is a pip or git type repo? I think we need two config values

server/src/inmantals/server.py
```diff
@@ -118,22 +125,83 @@ def flatten(self, line, char):
...
+        module_name: Optional[str] = None
+        if os.path.exists(v2_metadata_file):
+            mv2 = ModuleV2(project=None, path=self.rootPath)
+            module_name = mv2.name
+        elif os.path.exists(v1_metadata_file):
+            mv1 = ModuleV1(project=None, path=self.rootPath)
+            module_name = mv1.name
+
+        if not module_name:
+            raise ModuleNotFoundError
```
wouterdb, line 165: How does this get handled when there is no module and no project?

wouterdb (changes requested):

server/src/inmantals/server.py
Hugo-Inmanta: The setting is a yaml list e.g.: `[{'url': 'https://github.com/inmanta/', 'type': 'git'}, {'url': 'https://pypi.org/simple', 'type': 'package'}]`

Hugo-Inmanta: I followed the doc https://docs.inmanta.com/inmanta-service-orchestrator/latest/reference/projectyml.html#project-yml

server/src/inmantals/server.py
Hugo-Inmanta: I'll look into that

server/src/inmantals/server.py
Hugo-Inmanta: Added a custom exception + handle it as an error so that no trace is shown: 
`2023-03-01 11:43:46,763 __main__                 INFO    Starting language server
2023-03-01 11:43:46,766 inmantals.server         INFO    project created at /tmp/vs-code/project6025286518
2023-03-01 11:43:46,767 inmantals.server         ERROR   The Inmanta extension only works on projects and modules. Please make sure the current workspace is a valid project (see https://docs.inmanta.com/inmanta-service-orchestrator/latest/model_developers/project_creation.html) or module (see https://docs.inmanta.com/inmanta-service-orchestrator/latest/model_developers/module_creation.html). 
`

server/src/inmantals/server.py
wouterdb: oh, I would have never guessed , but I think that is good enough for now

server/src/inmantals/server.py
wouterdb: but is it reported to the end-user in some way?

Hugo-Inmanta: For now, the message above is only displayed in the `Inmanta Language Server` console output. I think I'll add a toast notification

wouterdb (approved):

Issue #1014, opened 2023-03-08
coloring highlighting is messed up in some cases


FloLey:
The color highlighting of string seems to be messed up.
For example :
```
s = "this is a string"
ss = 'this is also a string'
sss = """this is "still" a string"""
ssss = """Let's \"break" it"""
sssss = "aaaaaaaaaaaaaaaaaaah"
```
will be colored like this:
![image.png](https://images.zenhubusercontent.com/61a72a44b0dfd661a656b110/466d25c7-1061-477d-a19c-8648f0eb62fe)

sanderr:
The vim highlight rules seem to be all right, so if vscode uses a similar regex system (really no clue anymore if it does) we can easily copy them.

PR #1017, opened 2023-03-09
Issue/1014 coloring highlighting is messed up in some cases

Hugo-Inmanta:
closes https://github.com/inmanta/vscode-inmanta/issues/1014

Highlighting looks like this now:
Screenshot from 2023-03-09 16-40-14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changes in this release:
 - Add information about how to show the python interpreter information in the status bar and make it so the status bar should appear when opening an inmanta file.
 - A warning that suggests running the `inmanta project install` command is now shown when the compiler fails to install modules. (#894)
 - Add support to work on a module. (#891)
+- Fix string syntax highlighting. (#1014)
 
 # v 1.2.0 (2022-12-23)
 Changes in this release:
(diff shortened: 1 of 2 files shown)


sanderr (changes requested):
I can't go into depth now because I need to prioritize other things, but I noticed these rules (and the vim ones I mentioned as a basis in the ticket) are still slightly too naive. Consider the following snippet:
```
s = """""test"""" x = 1 " y = 2 "
std::print(s)
std::print(x)
```
Hard to read, but what it does is
1. define a string `s` with contents `""test"`
2. define a variable `x = 1`
3. a string with contents `y = 2`
4. print the value for both `s` and `x`. Don't print `y` because it doesn't exist, it's just in the string.

I updated my vim rules to highlight multi-line strings in orange and normal strings in yellow, in order to clearly show how it matches. The result is this:
screenshot_2023-03-10-105546

This gives the user the impression that the `x = 1` is just a string and `y = 2` is an actual statement, followed by the opening of a new string that is never closed. This is clearly not the case.

I updated the MLS rule to `syn region String start="\"\{3,5}" end="\"\{3,5}` which results in the following:
screenshot_2023-03-10-105557

This looks like the correct highlighting to me. I'll open a PR for the vim syntax changes, but if feasible (not sure if these support regexes) I think we should address these as well.

Hugo-Inmanta:
Now renders as such:
image (3)

PR #1026, opened 2023-03-13
Issue/991 cleanup old ls

FloLey:
This PR:
1.  makes sure there is only one instance of the logs of the LS use the same and cleans up the old ones
2. changes the condition to show the venv selector in the bottombar. It will now also be shown when the logs are open, a pip-requirements file is open, a config (walk through) is open.


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Changes in this release:
 - Add information about how to show the python interpreter information in the status bar and make it so the status bar should appear when opening an inmanta file.
 - A warning that suggests running the `inmanta project install` command is now shown when the compiler fails to install modules. (#894)
 - Add support to work on a module. (#891)
+- Fixes some unclear error messages (#970)
+- 'Export to server' is now run in terminal (#970)
+- Walkthrough/setup assistant added (#970)
+- Commands added to: install the language server, activate the language server, run 'project install' and to open the setup assistant (#970)
+- Fixed bug where the multiple outputs of the language server where visible (#991)
```
FloLey, line 11: I had forgotten to add this to my previous PR

src/language_server.ts
```diff
@@ -394,6 +394,9 @@ export class LanguageServer {
 				}
 				this.client = undefined;
 			}
+			if(this.lsOutputChannel){
+				this.lsOutputChannel.dispose();
```
sanderr, line 398: Does this mean we immediately clear the logs when we stop the server? Is this always desirable? Would it be feasible to just append to the old logs when we restart the server? Possibly with a clear separator.

src/python_extension.ts
```diff
@@ -75,7 +75,7 @@ export class PythonExtension {
 		}
 		const editor = window.activeTextEditor;
 		this.inmantaEnvSelector.text = text;
-		if (editor && editor.document.languageId === "inmanta") {
+		if (editor && ['inmanta','log','pip-requirements','properties'].includes(editor.document.languageId)) {
```
sanderr, line 78: This is unrelated, right? Does it still only show when you're working on an inmanta module/project, or does it now show for any project's `requirements.txt`?

sanderr (commented):
Some questions, no immediate change requests.

src/language_server.ts
```diff
@@ -26,6 +24,7 @@ enum LanguageServerDiagnoseResult {
 export class LanguageServer {
 	mutex = new Mutex();
 	client: LanguageClient;
+	lsOutputChannel:OutputChannel = null;
```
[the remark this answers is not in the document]
FloLey: I don't think so but I might be wrong.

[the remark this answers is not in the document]
sanderr: Looks like we don't, I'll create a ticket, should be easy to set up.

src/language_server.ts
FloLey: yes that's what it currently does. I though that was what we wanted. I'll try to see if I can make it append as I like it better to.

sanderr: I'm honestly not sure, I don't think we discussed the desired behavior in depth so I think we have some freedom here.

src/python_extension.ts
FloLey: Indeed, I noticed I forgot that in my previous pr to: 
It will now show when the active window is a .cf, vscode logs, a pip-requirements file or a config window (like the walk through).

sanderr: But still only if the active directory or one of the directories in the active workspace is an inmanta project/module? Or always for those file types, even for a plain Python project?

FloLey: First the extension needs to start ( currently you will still need to open a .cf file or to execute a command defined by the extension). After that our own env selector will start to show for those type of files.

src/language_server.ts
FloLey: video

Do you think this is enough of a separator?

sanderr: I think so: the line draws the eye, and then the "Started ..." and "Stopped ..." messages clearly show what it means. I like it.

sanderr (approved):

Hugo-Inmanta (approved):

PR #1029, opened 2023-03-14
Issue/891 fix support for modules

Hugo-Inmanta:
This PR fixes some issues with the support for modules:
- v2 modules were being loaded but not installed in the venv, they are now installed in editable mode
- a symlink to the v1 module is now added inside the libs folder of the temporary project

This PR also
- adds a Folder class which is a wrapper around a folder containing an inmanta project or an inmanta module opened in a vscode workspace
- moves the create_tmp_project method from the InmantaLSHandler to this Folder class
- adds some basic functionalities to interact with the folders settings


server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+
+    folder_uri: URI
+    name: str
+    inmanta_project_dir: Union[Optional[tempfile.TemporaryDirectory], str]
+    handler: "InmantaLSHandler"
+
+    def __init__(self, folder_uri: URI, name: str, handler: "InmantaLSHandler"):
+        self.folder_uri = folder_uri
+        self.name = name
+        self.handler = handler  # Keep a reference to the handler for cleanup
```
FloLey, line 87: Am I right to think it is with this you fixed the issue I reported yesterday?

server/src/inmantals/server.py
```diff
@@ -75,22 +216,57 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+        if workspaceFolders is None:
+            raise InvalidExtensionSetup("No workspace folder specified.")
 
-        if rootPath is None:
-            raise InvalidExtensionSetup("A folder should be opened instead of a file in order to use the inmanta extension.")
+        # Keep track of the folders opened in this workspace
+        self._workspace_folders: Dict[str, Folder] = Folder.unpack_workspaces(workspaceFolders, self)
+
+        if len(self._workspace_folders) > 1:
+            raise NotImplementedError("Working on multiple folders inside a workspace is not supported yet.")
```
FloLey, line 263: Do I understand correctly that this PR prepares everything that is needed to implement this but the rest is in the next PR? Some parts of the code seems to be written as if it can already happen that there is more than 1 _workspace_folders. (cleanup, sync_compile_and_anchor_folders). Shouldn't this be part of the next PR to instead of this one?

FloLey (approved):
I like it. I tried it a bit and it seems to work. Just a small question

server/src/inmantals/server.py
```diff
@@ -130,68 +312,29 @@ def flatten(self, line, char):
         assert char < 100000
         return line * 100000 + char
 
-    def create_tmp_project(self) -> str:
-        self.project_dir = tempfile.TemporaryDirectory()
```
Hugo-Inmanta, line 134: @FloLey here

server/src/inmantals/server.py
Hugo-Inmanta: Yes, the cleanup wasn't done right:
The create_tmp_project method used be part of the InmantaLSHandler class and create self.project_dir. The problem was that I moved this method to the Folder class, but I didn't change the cleanup logic(https://github.com/inmanta/vscode-inmanta/blob/662678c99471a8144c52f5a7e56e9aff8a1cb685/server/src/inmantals/server.py#L456-L458)

server/src/inmantals/server.py
Hugo-Inmanta: Yes exactly.
I agree this would be more clean to separate this in the next PR

FloLey: ok I don't really care. I just wanted to make sure I wasn't missing something :)

server/src/inmantals/server.py
```diff
@@ -75,22 +216,57 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+        if root_uri:
+            logger.warning("The rootUri parameter has been deprecated in favour of the 'workspaceFolders' parameter.")
+
+        if workspaceFolders is None:
+            raise InvalidExtensionSetup("No workspace folder specified.")
 
-        if rootPath is None:
-            raise InvalidExtensionSetup("A folder should be opened instead of a file in order to use the inmanta extension.")
+        # Keep track of the folders opened in this workspace
+        self._workspace_folders: Dict[str, Folder] = Folder.unpack_workspaces(workspaceFolders, self)
```
sanderr, line 260: I think it would be good to pull all input through `pydantic`, then we can also properly type this method. There are already a bunch of LSP types defined in `lsp_types.py`.

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+        try:
+            # Look for some custom settings for this folder
+            with open(os.path.join(self.get_folder_path(), ".vscode", "settings.json")) as settings_file:
+                settings: dict[str, str] = json.load(settings_file)
+                return settings.get(key, default)
+        except FileNotFoundError:
+            return default
+
+    @classmethod
+    def unpack_workspaces(cls, workspace_folders: Sequence[object], ls_handler: "InmantaLSHandler") -> Dict[str, "Folder"]:
```
sanderr, line 124: Since this seems to be one of the main entry points a docstring would be nice.

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+
+    def cleanup(self) -> None:
+        logger.info(f"calling cleanup for {self}")
+        if self.inmanta_project_dir:
+            self.handler.remove_folder(str(self.folder_uri))
+            self.inmanta_project_dir.cleanup()
+
+    def get_folder_path(self) -> str:
+        """
+        Convert the folders's uri into a regular path
```
sanderr, line 135:
```suggestion
        Convert the folder's uri into a regular path
```

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+                return settings.get(key, default)
+        except FileNotFoundError:
+            return default
+
+    @classmethod
+    def unpack_workspaces(cls, workspace_folders: Sequence[object], ls_handler: "InmantaLSHandler") -> Dict[str, "Folder"]:
+        return {folder["uri"]: Folder(URI(folder["uri"]), folder["name"], ls_handler) for folder in workspace_folders}
+
+    def cleanup(self) -> None:
+        logger.info(f"calling cleanup for {self}")
```
sanderr, line 128: Generally we shouldn't use f-strings for logging, recall Wouter's lightning talk at the last team week. There are some other logging instances where this should be applied, I didn't mark them all.
```suggestion
        logger.info("calling cleanup for %s", self)
```

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+                "install_mode": install_mode.value,
+            }
+            yaml.dump(metadata, fd)
+
+        with open(os.path.join(self.inmanta_project_dir.name, "main.cf"), "w+") as fd:
+            fd.write(f"import {module_name}\n")
+
+        return self.inmanta_project_dir.name
+
+    def __repr__(self):
```
sanderr, line 198: This is not what this method is for. It is a fine and likely not widely-known line so it's an easy mistake to make. Generally, `repr` is meant to be a structured, unambiguous, and ideally machine-consumable representation. `str` on the other hand is for human consumption and has no strict requirements. See the docs for [`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__) and `__str__` below it.

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+        Convert the folders's uri into a regular path
+        (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri)
+        """
+        return self.folder_uri.path
+
+    def create_tmp_project(self) -> str:
+        """
+        When working on a module, an inmanta project is created in a temporary directory.
+        """
+        self.inmanta_project_dir = tempfile.TemporaryDirectory()
```
sanderr, line 144: Do I understand correctly that the caller of `create_tmp_project` is expected to call `cleanup` at some later point? If so, it may be worth adding that invariant to the docstring. Could it also be useful to add a context manager that does the cleanup automatically?
```python
with Folder.tmp_project() as something:
    do_stuff(something)  # doesn't matter if this raises an exception because the context manager cleans up in a `finally` block
```

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+        if os.path.exists(v1_metadata_file):
+            mv1 = module.ModuleV1(project=None, path=self.get_folder_path())
+            module_name = mv1.name
+            os.symlink(self.get_folder_path(), os.path.join(libs_folder, module_name), target_is_directory=True)
+
+        elif os.path.exists(v2_metadata_file):
+            mv2 = module.ModuleV2(project=None, path=self.get_folder_path(), is_editable_install=True)
+            module_name = mv2.name
+            # Install this v2 module in editable mode in the active venv.
+            env.process_env.install_from_source([env.LocalPackagePath(path=self.get_folder_path(), editable=True)])
```
sanderr, line 166: Since this is just `inmanta.env`, which has no knowledge about project config or module sources, we need to make sure it uses the appropriate pip index(es). pytest-inmanta-lsm does something similar [here](https://github.com/inmanta/pytest-inmanta-lsm/blob/fffe3c9af9be030f525e3284aeaa277091f1ecc8/src/pytest_inmanta_lsm/resources/setup_project.py#L92). This case may or may not be slightly simpler because we have easier access to the configured repos.

Now that this is required for two separate things, it may also be worthwhile to add something like this to core. We probably couldn't use it here yet for backwards compatibility, but if you find yourself pretty much copying this thing from pytest-inmanta-lsm, could you create a ticket on core to make a method for this in core, and to update both this location and the pytest-inmanta-lsm one I linked to call the core method instead?

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+
+        if not module_name:
+            error_message: str = (
+                "The Inmanta extension only works on projects and modules. "
+                f"Please make sure the folder opened at {self.get_folder_path()} is a valid "
+                "[project](https://docs.inmanta.com/inmanta-service-orchestrator/latest/model_developers/project_creation.html)"
+                " or "
+                "[module](https://docs.inmanta.com/inmanta-service-orchestrator/latest/model_developers/module_creation.html)"
```
sanderr, line 174: This is a though one but I think we have to refer to the OSS docs (https://docs.inmanta.com/community/latest/) because not everyone can necessarily access these.

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+            )
+            self.cleanup()
```
sanderr, line 176: This one could be dropped if we go for the context manager approach.

server/src/inmantals/server.py
```diff
@@ -75,22 +216,57 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+        """Reads the default value for a setting set by the initializationOptions argument in the initialize method."""
+        logger.info(f"key={key} self.inmanta_settings={self.inmanta_settings}")
+        return self.inmanta_settings.get(key, default)
+
+    def remove_folder(self, folder_uri: str) -> None:
+        logger.info(self._workspace_folders)
+        del self._workspace_folders[folder_uri]
+
+    async def initialize(
+        self, workspaceFolders: Sequence[object], capabilities: Dict[str, object], root_path=None, root_uri=None, **kwargs
```
sanderr: Why the change `rootUri` -> `root_uri` and similar for `root_path`? Did something change in the protocol?

server/src/inmantals/server.py
```diff
@@ -75,22 +216,57 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+    ):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
+        logger.info(f"workspaceFolders={workspaceFolders}")
+
+        if root_path:
+            logger.warning("The rootPath parameter has been deprecated in favour of the 'workspaceFolders' parameter.")
+        if root_uri:
+            logger.warning("The rootUri parameter has been deprecated in favour of the 'workspaceFolders' parameter.")
```
sanderr, line 254: This one raised an exception before, why not anymore?

server/src/inmantals/server.py
```diff
@@ -75,22 +216,57 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
...
+
+    def remove_folder(self, folder_uri: str) -> None:
+        logger.info(self._workspace_folders)
+        del self._workspace_folders[folder_uri]
+
+    async def initialize(
+        self, workspaceFolders: Sequence[object], capabilities: Dict[str, object], root_path=None, root_uri=None, **kwargs
+    ):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
+        logger.info(f"workspaceFolders={workspaceFolders}")
```
sanderr, line 249: I think this would be more appropriate as debug.

server/src/inmantals/server.py
```diff
@@ -236,13 +379,18 @@ def treeify_reverse(iterator):
...
+            for folder in self._workspace_folders.values():
+                # run synchronous part in executor to allow context switching while awaiting
+                await asyncio.get_event_loop().run_in_executor(self.threadpool, sync_compile_and_anchor_folder, folder)
+                await self.publish_diagnostics(None)
+                logger.info(f"Compilation succeeded for folder {folder}")
+
         try:
             if self.shutdown_requested:
                 return
+            await sync_compile_and_anchor_folders()
```
sanderr, line 392: So we always compile all folders? I would think there would always be one active one and we'd only need to compile that one? If you think this question is better suited for the other PR that's fine as well.

server/src/inmantals/server.py
```diff
@@ -130,68 +312,29 @@ def flatten(self, line, char):
...
-                    project_dir: str = self.rootPath
-
+                if folder.inmanta_project_dir:
+                    project_dir: str = folder.get_project_dir()
+                    logger.info(f"using  existing project {project_dir}")
                 else:
-                    # Create a project in the vscode temp folder
-                    project_dir = self.create_tmp_project()
+                    # Create a temporary project
+                    project_dir = folder.create_tmp_project()
```
sanderr, line 330: memory leak? This creates a new project for each compile but it is never cleaned up. Either the method should be idempotent (and its docstring should mention it) or we should clean this up after each compile.

sanderr (changes requested):
Very clear and thorough description of the PR, I like it!

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+        v2_metadata_file: str = os.path.join(self.get_folder_path(), module.ModuleV2.MODULE_FILE)
+        v1_metadata_file: str = os.path.join(self.get_folder_path(), module.ModuleV1.MODULE_FILE)
+
+        module_name: Optional[str] = None
+        libs_folder = os.path.join(self.inmanta_project_dir.name, "libs")
+
+        if os.path.exists(v1_metadata_file):
+            mv1 = module.ModuleV1(project=None, path=self.get_folder_path())
+            module_name = mv1.name
+            os.symlink(self.get_folder_path(), os.path.join(libs_folder, module_name), target_is_directory=True)
```
sanderr, line 160: Open question: do we need to be compatible with any Windows machine? It's symlink support seems to have some limitations: https://docs.python.org/3/library/os.html#os.symlink.

package.json
```diff
@@ -59,13 +59,13 @@
                     "type": "string",
                     "default": "",
                     "description": "Path to the Python virtual environment used by the compiler. Will only be used by language server if inmanta-core<6. Leave empty to use the default. Only available when using Inmanta versions higher than 2020.5 in the language server environment.",
-                    "scope": "window"
+                    "scope": "resource"
```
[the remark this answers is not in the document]
Hugo-Inmanta: Here's the doc for reference: https://github.com/Microsoft/vscode/wiki/Adopting-Multi-Root-Workspace-APIs#settings
This is to allow different settings per folder in the workspace

server/src/inmantals/server.py
Hugo-Inmanta: I'll change all instances. I remember him giving a talk about this, but I don't recall the exact issue with this. I looked around a bit and found that for performance reasons the %s style is better. Was there another reason not to use f-strings ?

server/src/inmantals/server.py
Hugo-Inmanta: > Do I understand correctly that the caller of create_tmp_project is expected to call cleanup at some later point? 
Yes exactly. Here's how I see the flow:
When the server first realizes the open folder is a module -> create a tmp project that we should keep around until server shutdown
When a .cf file of the module is saved -> use the existing tmp project.

I considered using a context manager for this and use it inside the compile_and_anchor method but decided against it because it would mean  creating a new tmp project and tearing it down everytime this method is called. I figured it would be better to reuse the project created the first time around.

server/src/inmantals/server.py
Hugo-Inmanta: True, nice catch

package.json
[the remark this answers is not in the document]
sanderr: Thanks!

server/src/inmantals/server.py
sanderr: In general f-strings are fine, but the logging framework uses `%s` intrinsically (see how my suggestion has a comma rather than a `%`, so we really pass the values to the logging framework). There are some performance considerations, but more importantly it's just a good practice to do it as the logging framework intended. The worst scenario is if you would use a mixed style: `logger.info(f"hello {world} %s", "!")`. If `world` happens to contain another `%s`, this will result in an exception. So it's way safer than in some low level languages, but it's still something to take into account.

server/src/inmantals/server.py
sanderr: Ok, I have no objections against that. If not too much trouble it would be good to trigger the cleanup also when the server is killed or crashes. I don't know how much effort that would be.

server/src/inmantals/server.py
sanderr: I'll read the docs you linked and see if that changes anything for me, thanks.

server/src/inmantals/server.py
Hugo-Inmanta: Indeed, this is an artifact from the previous implementation. Only the "active" folder should get compiled.

server/src/inmantals/server.py
Hugo-Inmanta: Unless I'm missing an edge case there shouldn't be a memory leak. The flow is as following:

- During initialization the InmantaLSHandler [registers](https://github.com/inmanta/vscode-inmanta/blob/5e1fd05c03b10a4dc82915ab3591ffebf09bc80b/server/src/inmantals/server.py#L260) the active folder  (It used to register all folders, but it will be changed to only register the active folder)
- First compile: create a temporary project
- Following compiles: use the already created temporary project
- During shutdown: [cleanup](https://github.com/inmanta/vscode-inmanta/blob/5e1fd05c03b10a4dc82915ab3591ffebf09bc80b/server/src/inmantals/server.py#L451) all temporary projects (same, this used to cleanup all registered folders, but there will be only the active one soon)

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+        inmanta project used to load this module.
+        """
+        if not self.inmanta_project_dir:
+            return None
+        if isinstance(self.inmanta_project_dir, str):
+            return self.inmanta_project_dir
+        else:
+            return self.inmanta_project_dir.name
+
+    def get_setting(self, key: str, default: Optional[str] = None) -> Optional[str]:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Yes this feels hackish to me as well, I think I can find a way around this now that the refactor of the language server has been merged

server/src/inmantals/server.py
sanderr: Ah I see, I overlooked the third bullet point. I withdraw this comment.

server/src/inmantals/server.py
```diff
@@ -64,9 +65,149 @@ def __init__(self, message: str) -> None:
...
+    Wrapper class around a folder (inmanta module or inmanta project). This folder can either be standalone or live in a
+    workspace alongside other folders.
+
+    The constructor should not be called explicitly. Instantiation is done through the unpack_workspaces method
+    :param folder_uri: uri of the folder that is assumed to live in a workspace
+    :param name: name of the folder
+    :param handler: reference to the InmantaLSHandler responsible for this folder
+    """
+
+    folder_uri: URI
```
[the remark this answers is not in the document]
Hugo-Inmanta: I think I can do without

server/src/inmantals/server.py
Hugo-Inmanta: My mistake, got reverted

server/src/inmantals/server.py
Hugo-Inmanta: I forgot to address this. I could catch OSError and suggest the user activates  Developer Mode  or runs vscode as admin ?

sanderr: I just asked Bart. Apparently we are already incompatible with Windows so we can leave it as is.

FloLey (approved):

src/language_server.ts
```diff
@@ -360,7 +378,7 @@ export class LanguageServer {
 	/**
 	 * Starts or restarts the language server.
 	 * @param {boolean} start Whether to start the server or restart it.
-	 * @returns {Promise<void>}
+	 * @returns {Promise<void>}.
```
FloLey, line 381:
```suggestion
	 * @returns {Promise<void>}
```

FloLey (approved):

server/src/inmantals/server.py
```diff
@@ -76,110 +79,145 @@ class Folder:
...
-        """
-        Convert the folders's uri into a regular path
-        (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri)
-        """
-        return self.folder_uri.path
+    def install_v2_module_editable_mode(self):
+        # TODO: this method is heavily inspired by https://github.com/inmanta/pytest-inmanta-lsm/blob/fffe3c9af9be030f525e3284aeaa277091f1ecc8/src/pytest_inmanta_lsm/resources/setup_project.py#L92  # NOQA E501
+        # To avoid code duplication, this should be made into a method into core and we should call this method here and in
+        # pytest-inmanta-lsm
+        # TODO: add ticket reference once its created ?
```
sanderr, line 125: Yes, sounds good!

server/src/inmantals/server.py
```diff
@@ -76,110 +79,145 @@ class Folder:
...
-            self.handler.remove_folder(str(self.folder_uri))
             self.inmanta_project_dir.cleanup()
 
-    def get_folder_path(self) -> str:
-        """
-        Convert the folders's uri into a regular path
-        (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri)
-        """
-        return self.folder_uri.path
+    def install_v2_module_editable_mode(self):
```
sanderr, line 121: Could you add a docstring that declares the precondition: this method assumes that the project is already set up (`module.Project.set()`), including the module repos config.

server/src/inmantals/server.py
```diff
@@ -57,41 +60,252 @@
...
+            with open(os.path.join(self.folder_uri, ".vscode", "settings.json")) as settings_file:
+                settings: dict[str, str] = json.load(settings_file)
+                return settings.get(key, default)
+        except FileNotFoundError:
+            return default
+
+    def cleanup(self) -> None:
+        logger.info("Cleaning up %s...", self)
+        if self.inmanta_project_dir:
+            self.inmanta_project_dir.cleanup()
```
sanderr, line 119: `self.inmanta_project_dir` could be just a string, no? Is this method no longer meant to be used now that you introduced `register_tmp_project`?

server/src/inmantals/server.py
```diff
@@ -76,110 +79,145 @@ class Folder:
...
-        If this folder holds an inmanta module, this returns the path of the temporary
-        inmanta project used to load this module.
-        """
-        if not self.inmanta_project_dir:
-            return None
-        if isinstance(self.inmanta_project_dir, str):
-            return self.inmanta_project_dir
-        else:
-            return self.inmanta_project_dir.name
+            self.inmanta_project_dir = self.create_tmp_project()
```
sanderr, line 100:
```suggestion
            # self.kind is set by this method
            self.inmanta_project_dir = self.create_tmp_project()
```

server/src/inmantals/server.py
```diff
@@ -190,84 +228,84 @@ def create_tmp_project(self) -> str:
...
-            self.add_setting("inmanta.compiler_venv", init_options.get("compilerVenv", None))
-            self.add_setting("inmanta.repos", init_options.get("repos", None))
+            self.compiler_venv_path = init_options.get(
+                "compilerVenv", os.path.join(os.path.abspath(urlparse(workspaceFolders[0]["uri"]).path), ".env-ls-compiler")
+            )
+            self.repos = init_options.get("repos", None)
+            logger.debug("self.repos= %s", self.repos)
 
+        # Keep track of the root folder opened in this workspace
+        self.root_folder: Folder = Folder(workspaceFolders[0], self)
```
sanderr, line 308: You defined the pydantic type but you're not using it. Ideally you'd do the validation a bit further up, so that you can also use the validated object for the compiler venv.

server/src/inmantals/server.py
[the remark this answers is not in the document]
sanderr: Looks a lot better now. I think you can remove this method now?

package.json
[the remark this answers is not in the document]
sanderr: Ok, looks good! Could you add the link in a comment?

server/src/inmantals/server.py
```diff
@@ -57,41 +60,252 @@
...
         Exception.__init__(self, message)
         self.message = message
 
 
+class Folder:
+    """
+    Wrapper class around a folder (inmanta module or inmanta project). This folder can either be standalone or live in a
+    workspace alongside other folders.
+
+    The constructor should not be called explicitly. Instantiation is done through the unpack_workspaces method
```
sanderr, line 76: I think this is no longer accurate.

server/src/inmantals/server.py
sanderr: Is this part something you cover in the other PR? "Since workspace folders can come and go dynamically the server also needs to register for workspace/didChangeWorkspaceFolders notifications."

server/src/inmantals/server.py
```diff
@@ -76,110 +79,145 @@ class Folder:
...
+    folder_uri: str
     name: str
     inmanta_project_dir: Union[Optional[tempfile.TemporaryDirectory], str]
     handler: "InmantaLSHandler"
 
-    def __init__(self, folder_uri: URI, name: str, handler: "InmantaLSHandler"):
-        self.folder_uri = folder_uri
-        self.name = name
+    def __init__(self, workspace_folder: lsp_types.WorkspaceFolder, handler: "InmantaLSHandler"):
+        folder_uri = urlparse(workspace_folder["uri"])
```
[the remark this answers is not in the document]
Hugo-Inmanta: I don't think so, because of [this](https://github.com/inmanta/vscode-inmanta/blob/aded8301d20e70376857c0577a4c7ec84fb55650/src/language_server.ts#L225) the language server should only react to files on disk.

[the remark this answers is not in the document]
sanderr: Cool, then we could even use pydantic's `FileUrl` type for validation.

server/src/inmantals/server.py
```diff
@@ -76,110 +79,145 @@ class Folder:
...
+
+        self.folder_uri = os.path.abspath(folder_uri.path)
         self.handler = handler  # Keep a reference to the handler for cleanup
+        self.kind: object
```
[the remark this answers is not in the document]
Hugo-Inmanta: Didn't know about this, it looks perfect for what I needed :+1:

server/src/inmantals/server.py
Hugo-Inmanta: > Is this part something you cover in the other PR
Yes, I ended up using the rootUri after all in the other PR

server/src/inmantals/server.py
```diff
@@ -598,13 +623,13 @@ async def dispatch_method(self, id, method, params):
...
         if pym is None:
-            logger.debug("Call to unexisting method %s params %s", method, params)
+            # logger.debug("Call to unexisting method %s params %s", method, params)
             raise MethodNotFoundException("Could not find method %s" % method, id)
 
         if params is None:
             params = {}
 
-        logger.debug("Called: %s %s", pymethod, params)
+        # logger.debug("Called: %s %s", pymethod, params)
```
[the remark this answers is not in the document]
Hugo-Inmanta: I forgot I removed those while debuging, adding them back in.

package.json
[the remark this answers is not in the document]
Hugo-Inmanta: Added a comment in the __init__ of InmantaLSHandler since I think comments aren't allowed in json files ?

server/src/inmantals/server.py
Hugo-Inmanta: > self.inmanta_project_dir could be just a string, no?

Yes that would be much cleaner :+1: 

> Is this method no longer meant to be used now that you introduced register_tmp_project?

Indeed, removing it

server/src/inmantals/server.py
```diff
@@ -290,6 +276,10 @@ async def initialize(
             if rootUri is None:
                 raise InvalidExtensionSetup("No workspace folder or rootUri specified.")
             workspaceFolders = [rootUri]
+            workspace_folder = lsp_types.WorkspaceFolder(uri=rootUri, name=os.path.dirname(urlparse(rootUri).path))
+        else:
+            workspace_folder = lsp_types.WorkspaceFolder(uri=workspaceFolders[0]["uri"], name=workspaceFolders[0]["name"])
```
sanderr, line 281: If the data is invalid this might fail with a `KeyError` so it's better to pass it as a whole to `pydantic`:
```suggestion
            workspace_folder = lsp_types.WorkspaceFolder(**workspaceFolders[0]["uri"])
```

The one using `rootUri` already is the best we can do I think.

package.json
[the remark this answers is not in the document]
sanderr: I thought they were but I could be wrong. The comment you added is fine.

server/src/inmantals/server.py
sanderr: I mean the part I quoted about `workspace/didChangeWorkspaceFolders`. It's missing here but if it's part of the other PR that's all right. So is it?

server/tests/test_smoke.py
```diff
@@ -29,11 +29,11 @@
...
 from inmantals import lsp_types
 from inmantals.jsonrpc import JsonRpcServer
-from inmantals.server import InmantaLSHandler
+from inmantals.server import CORE_VERSION, InmantaLSHandler
 from packaging import version
-from pkg_resources import Requirement, get_distribution, parse_requirements
+from pkg_resources import Requirement, parse_requirements
 
-INMANTA_CORE_VERSION = version.Version(get_distribution("inmanta-core").version)
+# CORE_VERSION = version.Version(get_distribution("inmanta-core").version)
```
sanderr, line 36: This can go now?

sanderr (approved):

server/src/inmantals/server.py
Hugo-Inmanta: Yes it is handled in the other PR, although not directly at the language server level, since each ls is only responsible for one folder in the workspace, so in a sense they don't care about the changes to the workspace.
It is handled in the extension, namely keeping track of each top level opened folder and its associated ls, and listening to changes to the workspace and starting/stopping the right ls when a folder is added/removed to the workspace.

sanderr: I see. That probably suffices. In any case, I'll consider that part of the other PR's review.

sanderr (approved):

PR #1035, opened 2023-03-16
Docstring on hover and pop-up for constructor/plugin definition

FloLey:
support hovering an entity or a plugin definition to view its docstring
Peek 2023-03-17 10-42


server/src/inmantals/server.py
```diff
@@ -53,8 +54,30 @@
 Recent versions use the encapsulating environment and require explicit project installation as a safeguard.
 """
 
-logger = logging.getLogger(__name__)
+BEFORE_ANCHOR_TARGET: bool = CORE_VERSION <= version.Version("8.3.dev")
```
FloLey, line 57: I think 8.3 is right but I'm not entirely sure. It is the version of my master branch of core.

FloLey:
(The tests wont pass as long as the core PR hasn't been merged)

server/src/inmantals/server.py
wouterdb: You could also do the import and catch the exception, that is a very robust way of detecting it.

server/src/inmantals/server.py
```diff
@@ -210,6 +234,7 @@ def setup_project():
             (statements, blocks) = compiler_instance.compile()
             scheduler_instance = scheduler.Scheduler()
             anchormap = scheduler_instance.anchormap(compiler_instance, statements, blocks)
+            anchermap_with_anchorTarget = [(s, AnchorTarget(t)) if isinstance(t, Range) or isinstance(t, Location) else (s,t)  for s, t in anchormap]
```
wouterdb, line 237:
```suggestion
            # Make sure everything is an AnchorTarget, this is for backward compatibility
            anchermap_with_anchorTarget = [(s, AnchorTarget(t)) if isinstance(t, Range) or isinstance(t, Location) else (s,t)  for s, t in anchormap]
```

server/src/inmantals/server.py
```diff
@@ -321,25 +345,41 @@ async def textDocument_didSave(self, **kwargs):  # noqa: N802
     async def textDocument_didClose(self, **kwargs):  # noqa: N802
         pass
 
-    def convert_location(self, loc):
+    def convert_location(self, target):
         prefix = "file:///" if os.name == "nt" else "file://"
-        if isinstance(loc, Range):
+        if isinstance(target, AnchorTarget) and isinstance(target.location, Range):
```
wouterdb, line 350: Why is there the `isinstance(target, AnchorTarget)` 

I assume all locations are pre-converted? 
if not, the else branch will still be broken I think? (e.g. assuming `target` is a` Range`)

server/src/inmantals/server.py
```diff
@@ -321,25 +345,41 @@ async def textDocument_didSave(self, **kwargs):  # noqa: N802
...
+
+    def get_file_type(self, filepath: str) -> str:
+        file_extension = os.path.splitext(filepath)[1].lower()
+        if file_extension == ".py":
+            return "python"
+        elif file_extension == ".cf":
+            return "inmanta"
+        else:
+            return ""
+
```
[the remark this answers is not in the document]
wouterdb: I'm not sure that would be better. 

1. it would tangle the two code bases more
2. it is a simple operation

server/src/inmantals/server.py
```diff
@@ -380,6 +420,43 @@ async def textDocument_references(self, textDocument, position, context):  # noq
 
         return [self.convert_location(loc.data) for loc in range]
 
+    async def textDocument_hover(self, textDocument, position):
+        logger.warn(position)
```
wouterdb, line 424:
```suggestion
```

server/src/inmantals/server.py
```diff
@@ -380,6 +420,43 @@ async def textDocument_references(self, textDocument, position, context):  # noq
...
+
+        if url not in self.anchormap:
+            return {}
+
+        tree = self.anchormap[url]
+
+        range = tree[self.flatten(position["line"], position["character"])]
+
+        if range is None or len(range) == 0:
+            return {}
```
wouterdb, line 440: I think this could be a helper method, as it is done for different endpoints

wouterdb (approved):
nice, I have some minor comments

server/src/inmantals/server.py
FloLey: Ow I forgot to clean this up, this is form the implementation before I noticed it was not backward compatible. all locations are pre-converted indeed

server/src/inmantals/server.py
```diff
@@ -321,64 +346,96 @@ async def textDocument_didSave(self, **kwargs):  # noqa: N802
...
+        if not range:
             return {}
 
-        return [self.convert_location(loc.data) for loc in range]
+        data = list(range)[0].data
+        logger.warn(list(range)[0])
+        docstring = textwrap.dedent(data.docstring.strip("\n")) if data.docstring else ""
+        docstring = docstring.replace(" ", "&nbsp;")
+        definition = self.get_definition(data).strip()
+        language = self.get_file_type(data.location.file)
```
sanderr, line 426: We don't delegate to the Python extension for `.py` files? Could we?

sanderr (changes requested):
Could you create a new ticket for the on-call docstring and put it in the stage 2 epic? It was initially part of this one but if I understand correctly it wouldn't be trivial?

server/src/inmantals/server.py
```diff
@@ -220,20 +246,19 @@ def treeify(iterator):
...
-            self.anchormap = {os.path.realpath(k): treeify(v) for k, v in groupby(anchormap, lambda x: x[0].file)}
-
+            self.anchormap = {os.path.realpath(k): treeify(v) for k, v in groupby(anchermap_with_anchorTarget, lambda x: x[0].file)}
             def treeify_reverse(iterator):
                 tree = IntervalTree()
                 for f, t in iterator:
-                    if isinstance(t, Range):
-                        start = self.flatten(t.lnr - 1, t.start_char - 1)
-                        end = self.flatten(t.end_lnr - 1, t.end_char - 1)
+                    if isinstance(t, AnchorTarget) and isinstance(t.location, Range):
```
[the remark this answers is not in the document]
FloLey: indeed, I changed it. This was some left over from my previous solution.

FloLey:
> Could you create a new ticket for the on-call docstring and put it in the stage 2 epic? It was initially part of this one but if I understand correctly it wouldn't be trivial?

https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/vscode-inmanta/1044

server/src/inmantals/server.py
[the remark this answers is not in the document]
FloLey: I will leave it as is for now but maybe this is something to keep in mind fro this ticket:
https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/gh/inmanta/vscode-inmanta/1044

server/src/inmantals/server.py
```diff
@@ -102,7 +105,7 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
 
     async def initialize(self, rootPath, rootUri, capabilities: Dict[str, object], **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
-
+        Config.set("compiler", "cache", "False")
```
sanderr, line 108: I don't think we want this in general, just for testing. So you need to set the corresponding env var in the Jenkinsfile.

server/src/inmantals/server.py
```diff
@@ -399,27 +405,44 @@ def get_range_from_position(self, textDocument, position)-> Optional[Interval]:
             return None
         return range
 
+    def get_range_from_position_reverse(self, textDocument, position) -> Optional[Interval]:
```
sanderr, line 408: There's a lot of duplication between this method and the one above. I think it could be simplified a lot by extracting this into a separate method that just accepts the dict to do the lookup in.

sanderr (changes requested):

src/test/runTest.ts
```diff
@@ -10,8 +10,10 @@ async function main() {
 	const tmpHomeDir: string = fs.mkdtempSync("/tmp/vscode-tests");
 	try {
 		const settings = {
+			// eslint-disable-next-line @typescript-eslint/naming-convention
 			"inmanta.ls.enabled": true,
-			"python.defaultInterpreterPath": process.env.INMANTA_EXTENSION_TEST_ENV
+			// eslint-disable-next-line @typescript-eslint/naming-convention
+			"python.defaultInterpreterPath": process.env.INMANTA_EXTENSION_TEST_ENV // eslint-disable-next-line camelcase
```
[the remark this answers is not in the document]
FloLey: the warning we get is: `Object Literal Property name `inmanta.ls.enabled` must match one of the following formats: camelCase`
I think dots are not allowed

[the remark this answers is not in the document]
sanderr: But that's horrible because this is a plain string, this could be literally anything, couldn't it? Or does it somehow recognize the context we're in?

src/test/runTest.ts
```diff
@@ -35,7 +35,8 @@ async function main() {
 		// in the home dir.
 		const extensionTestsEnv = {
 			HOME: tmpHomeDir,  // eslint-disable-line @typescript-eslint/naming-convention
-			INMANTA_LANGUAGE_SERVER_PATH: process.env.INMANTA_LANGUAGE_SERVER_PATH  // eslint-disable-line @typescript-eslint/naming-convention
+			INMANTA_LANGUAGE_SERVER_PATH: process.env.INMANTA_LANGUAGE_SERVER_PATH,  // eslint-disable-line @typescript-eslint/naming-convention,
+			INMANTA_COMPILER_CACHE: "False" // eslint-disable-line @typescript-eslint/naming-convention,
```
sanderr, line 39: Even better than the Jenkinsfile.

sanderr (approved):

src/test/runTest.ts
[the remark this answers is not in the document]
FloLey: I don't like it neither. At first I didn't want to add it but I got annoyed of seeing the warning each time when I know nothing is wrong.

src/test/runTest.ts
FloLey: But it seems it doesn't work :(

src/test/runTest.ts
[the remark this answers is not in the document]
sanderr: Looks like JS doesn't really have a separate map type, just objects. Therefore the rules seem to be muddled a bit, because generally (afaik) objects are used a lot like in Python, in which case its keys are identifiers, and should in fact match naming conventions. I can see how this would make linting very difficult. I found this related ticket: typescript-eslint/typescript-eslint#1483. I think it contains some workarounds but I didn't dive too deeply myself.

src/test/runTest.ts
sanderr: The LS tests run against multiple versions of the product and should be compatible with all of them, so it shouldn't be required there.

sanderr: (I added a comment that didn't make sense and deleted it right after. You probably got a notification, perhaps I should have kept it)

FloLey: No issue, I've done the same multiple times already :p

server/src/inmantals/server.py
```diff
@@ -77,9 +103,10 @@ def __init__(self, instream: BaseIOStream, outstream: BaseIOStream, address):
         # compiler_venv_path is only relevant for versions of core that require a compiler venv. It is ignored otherwise.
         self.compiler_venv_path: Optional[str] = None
 
-    async def initialize(self, rootPath, rootUri, capabilities: Dict[str, object], **kwargs):  # noqa: N803
+    async def initialize(self, rootPath, rootUri, capabilities: Dict[str, object], useCache=True, **kwargs):  # noqa: N803
         logger.debug("Init: " + json.dumps(kwargs))
-
+        if not useCache:
+            Config.set("compiler", "cache", "False")
```
sanderr, line 109: It may be simpler to just pass `Project(attach_cf_cache=useCache)`. You'd also need to check when the cache was introduced and whether we need to do some backward compatibility tricks.

Side note: we use camelCase for these options here because that's how they're defined in the spec. For our own options I'd go for snake_case.

server/src/inmantals/server.py
```diff
@@ -195,7 +220,7 @@ def setup_project():
                     else:
                         module.Project.set(module.Project(project_dir))
                 else:
-                    module.Project.set(module.Project(project_dir))
+                    module.Project.set(module.Project(project_dir, attach_cf_cache=useCache))
```
FloLey, line 223: @sanderr is this ok? 
I don't really like this part
`os.getenv("ENV_VAR", "False").lower() in ("true", "1")` but is seems I always get a string from os.getenv

Issue #892, opened 2022-12-21
Support workspaces


wouterdb:
Workspaces can have multiple root folders

Make sure this it is supported to have multiple projects/modules in one workspace

sanderr:
This may be relevant: https://github.com/Microsoft/vscode/wiki/Adopting-Multi-Root-Workspace-APIs#language-client--language-server

This applies to us: "If your extension is making use of the (now deprecated) workspace.rootPath property to work on the currently opened folder, then you are affected. See the section 'Eliminating workspace.rootPath below'."

PR #1061, opened 2023-03-24
Issue/892 add support for workspaces

Hugo-Inmanta:
Closes https://github.com/inmanta/vscode-inmanta/issues/892

In this PR:
- move the LsErrorHandler from extension.ts to language_server.ts
- Add the languageServers map to extension.ts that keeps track of currently active language servers
- Adds a more user-friendly way of selecting the repos for each folder in the workspace
- Any time a different folder is selected, all the inmanta commands are re-registered

Link to the ADR explaining why we went with one language server per folder: https://github.com/inmanta/inmanta-core/pull/5756


src/python_extension.ts
```diff
@@ -114,10 +146,12 @@ export class PythonExtension {
 		pythonApi.settings.onDidChangeExecutionDetails(
 			(resource: Resource) => {
 				let newExecutionDetails = pythonApi.settings.getExecutionDetails(resource);
+				let folder = workspace.getWorkspaceFolder(resource);
+				let outermost = getOuterMostWorkspaceFolder(folder).uri;
```
FloLey, line 150: Why isn't this done in the updatePythonPath function so it doesn't need to be passed as argument in the callback? I'm sad it needs arguments as this is ugly :p But I don't see an easy way around it.

src/python_extension.ts
```diff
@@ -84,28 +111,33 @@ export class PythonExtension {
...
 	}
 
 
 	/**
 	 * register a function that will be called when the "onChange" function is called
-	 * @param {() => void} onChangeCallback A function
+	 * @param {(newPath?) => void} onChangeCallback A function
 	 */
-	registerCallbackOnChange(onChangeCallback: () => void) {
+	registerCallbackOnChange(onChangeCallback: (newPath?, outermost?) => void, tags?) {
```
FloLey, line 129: the typing should be done like this but I'm not sure why `tags` is needed? 
```suggestion
	registerCallbackOnChange(onChangeCallback: (newPath?, outermost?) => void | tags) {
```

FloLey (commented):
Well played, there was a lot of work :p
most of the change requests are personal preferences. 
There might be some docstrings missing as I only noticed at the end of the review I didn't pay attention to.

server/src/inmantals/server.py
```diff
@@ -224,18 +224,20 @@ def create_tmp_project(self) -> str:
...
+            "name": "Temporary project",
+            "description": "Temporary project",
+            "modulepath": "libs",
+            "downloadpath": "libs",
+            "install_mode": install_mode.value,
+        }
+        if self.handler.repos:
+            metadata["repo"] = self.handler.repos
+        logger.debug(
+            "project.yaml created at %s, repos=%s", os.path.join(inmanta_project_dir, "project.yml"), self.handler.repos
```
sanderr, line 237: Do you know how it comes that many log messages are only in this file and not in the output?

server/src/inmantals/server.py
```diff
@@ -311,16 +314,17 @@ async def initialize(
             )
 
         init_options = kwargs.get("initializationOptions", None)
+        logger.debug("init_options= %s", init_options)
 
         if init_options:
             self.compiler_venv_path = init_options.get(
-                "compilerVenv", os.path.join(os.path.abspath(urlparse(workspace_folder.uri).path), ".env-ls-compiler")
+                "compilerVenv", os.path.join(os.path.abspath(urlparse(rootUri).path), ".env-ls-compiler")
```
sanderr, line 321: I don't get this change.

server/src/inmantals/server.py
```diff
@@ -351,12 +355,6 @@ def to_symbol_kind(value: int) -> Optional[lsp_types.SymbolKind]:
                     # the language server does not report work done progress for workspace symbol requests
                     "workDoneProgress": False,
                 },
-                "workspace": {
-                    "workspaceFolders": {
-                        "supported": True,
-                        "changeNotifications": True,
-                    }
-                },
```
sanderr, line 359: I don't get this change either.

sanderr (commented):
First review stage: reviewed server side only, excluding its tests.

server/src/inmantals/server.py
Hugo-Inmanta: As far as I can tell this is only useful when a language server is responsible for multiple root folders in the workspace.
This should never be the case here, as each language server should only be responsible for only one root folder, so in a sense they don't care about changes to other folders in the workspace.
The extension on the other hand is doing all the matching of root folders to language servers and is listening to changes to the workspace folders to start and stop language servers as required.
I drew inspiration from [The multi language server example](https://github.com/Microsoft/vscode-extension-samples/tree/main/lsp-multi-server-sample)

package.json
```diff
@@ -59,14 +59,48 @@
...
+                                  "Url of the git repository containing a v1 module",
+                                  "Pip index url in which to look for v2 modules"
+                                ],
+                                "order": 2
+                            }
+                        }
+                    },
+                    "default": [
+                        {"url": "https://github.com/inmanta/", "type": "git"},
+                        {"url": "https://pypi.org/simple/", "type": "package"}
```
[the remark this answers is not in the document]
Hugo-Inmanta: Making it part of the setup wizard is a very good idea, and I'm also in favor of erring towards the safer option and dropping it.

src/extension.ts
```diff
@@ -1,106 +1,257 @@
...
+		registerCommands(languageServer);
+	}
 
-	// Subscribe to workspace configuration changes and restart the language server if necessary
-	context.subscriptions.push(workspace.onDidChangeConfiguration(async event => {
-		if (event.affectsConfiguration('inmanta')) {
-			await languageserver.startOrRestartLS();
+	async function didOpenTextDocument(document: TextDocument): Promise<void> {
+		// We are only interested in .cf files
+		if (document.languageId !== 'inmanta' || (document.uri.scheme !== 'file' && document.uri.scheme !== 'untitled')) {
```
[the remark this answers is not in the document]
Hugo-Inmanta: This makes sure the opened document is a .cf file, and that it is either a file on disk or a newly created file that hasn't been saved yet (For reference other possible schemes: https://code.visualstudio.com/api/references/document-selector#summary).
Now that I think about it I think I'll remove the support for unsaved files

src/language_server.ts
```diff
@@ -419,4 +480,14 @@ export class LanguageServer {
 			}
 		});
 	}
+
+	cleanOutputChannel() {
```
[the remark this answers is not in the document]
Hugo-Inmanta: > If I understand correctly this will remove the logs of each LS whenever we switch workspace?

I'm not 100% what you mean exactly by "switching workspace", but here's how it works at the moment:

We only remove the logs of the ls whose folders have been removed from the workspace. e.g. you're working on a workspace with 3 folders: a project and 2 modules. Assuming they are all distinct folders (not parent of each other), then you would have 3 language servers, one per folder with the name of the folder as log title.

The cleanup would only happen if you removed one folder from the workspace (e.g. you remove the project folder, then you would only have left the logs for the 2 modules)

src/python_extension.ts
Hugo-Inmanta: I agree this is ugly, I got it to work that way and kept it at that since I'm not very comfortable with typescript callbacks, I'll have another quick look

package.json
```diff
@@ -62,9 +62,32 @@
                     "scope": "resource"
                 },
                 "inmanta.repos": {
-                    "type": "string",
-                    "default": "",
-                    "markdownDescription": "Optional yaml list of repositories to indicate the Inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories defined in the repo section of the [project.yaml](https://docs.inmanta.com/community/latest/reference/projectyml.html#project-yml) file.",
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "title": "inner objects",
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, it was indeed visible when hovering over the brackets in settings.json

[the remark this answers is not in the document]
Hugo-Inmanta: Will fix it when improving the wording of the whole thing

server/src/inmantals/server.py
Hugo-Inmanta: Nice catch

server/src/inmantals/server.py
Hugo-Inmanta: Not sure if this is what you mean but changing the log level to debug in there https://github.com/inmanta/vscode-inmanta/blob/1cbc27efec7136a0c2096372f906a30391c1ee2a/server/src/inmantals/jsonrpc.py#L141 shows these debug logs in the  language server output. (btw This is a "debug tips and tricks" I'm adding to the docs )

src/language_server.ts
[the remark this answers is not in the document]
FloLey: ok that's nice! :)

FloLey (approved):

src/extension.ts
```diff
@@ -1,106 +1,213 @@
...
+			const stdout = spawnResult.stdout.toString();
+			if (spawnResult.status === 1) {
+				const response = await window.showErrorMessage(`The active python interpreter is not part of a virtual environment.`, 'Select interpreter');
+				if (response === 'Select interpreter') {
+					await commands.executeCommand('python.setInterpreter');
+				}
+				else {
+					return;
+				}
+			}
```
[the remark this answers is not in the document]
FloLey: I'm not sure I understand what you mean :p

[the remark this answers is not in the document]
Hugo-Inmanta: This check makes sure that the python interpreter chosen by the user is actually a venv ( and not a global interpreter like `/bin/python` for example) so that when we don't install module dependencies in the global context of the user. (`context` might not be the correct terminology :thinking: ).

I have now moved this check back to the canServerStart() method.

media/markdown_set_module_dependency.md
```diff
@@ -0,0 +1,23 @@
...
+These module dependencies follow this scheme:
+
+```json
+    {"type": "git", "url": "https://github.com/inmanta/"}
+```
+
+The type property can have two values:
+# Type "git"
+
+The url property holds the URI of the parent location for a v1 module. (Organization url for github, Group url for gitlab, parent folder for a local repository...)
```
sanderr, line 17: Nicely worded!

media/markdown_set_module_dependency.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+# Type "package"
+
+
+The url property holds the pip index url in which to look for v2 modules. ⚠️ Be mindful of [dependency confusion attacks](https://www.websecuritylens.org/how-dependency-confusion-attack-works-and-how-to-prevent-it/) when setting this up: make sure you have full control over this index.
```
sanderr, line 23: I think we can be a bit less alarming here. It's good to point this out, but let's make sure to make clear it's only really an issue if you use more than one index. If we want to keep it short we could just say there are potential security risks when setting more than one index and point to https://docs.inmanta.com/community/dev/reference/projectyml.html#inmanta.module.ProjectMetadata for more information. I have no strong preference between that and just expanding on this to make it more clear.

package.json
```diff
@@ -65,29 +65,30 @@
...
                             },
                             "type":{
                                 "type": "string",
                                 "enum": ["git", "package"],
-                                "enumDescriptions": [
-                                  "Url of the git repository containing a v1 module",
-                                  "Pip index url in which to look for v2 modules"
+                                "markdownEnumDescriptions": [
+                                  "URI of the parent location for a v1 module. (Organization url for github, Group url for gitlab, parent folder for a local repository...)",
+                                  "Pip index url in which to look for v2 modules. ⚠️ Be mindful of [dependency confusion attacks](https://www.websecuritylens.org/how-dependency-confusion-attack-works-and-how-to-prevent-it/) when setting this up."
```
sanderr, line 80: Same here

package.json
```diff
@@ -65,29 +65,30 @@
...
-                        {"url": "https://github.com/inmanta/", "type": "git"},
-                        {"url": "https://pypi.org/simple/", "type": "package"}
+                        {"url": "https://github.com/inmanta/", "type": "git"}
                     ],
-                    "markdownDescription": "List of git repositories and pip indexes to indicate the Inmanta extension where to look for modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories and pip indexes defined in the repo section of the [project.yaml](https://docs.inmanta.com/community/latest/reference/projectyml.html#project-yml) file.",
+                    "markdownDescription": "List of dependency module locations to indicate the Inmanta extension where to look for dependency modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories and pip indexes defined in the repo section of the [project.yaml](https://docs.inmanta.com/community/latest/reference/projectyml.html#project-yml) file.",
```
sanderr, line 91:
```suggestion
                    "markdownDescription": "List of dependency module locations to indicate the Inmanta extension where to look for dependency modules when working on an individual module.\nWhen working on a project, this option is ignored and the Inmanta extension will look for modules in the repositories and pip indexes defined in the repo section of the [project.yml](https://docs.inmanta.com/community/latest/reference/projectyml.html#project-yml) file.",
```

server/src/inmantals/server.py
sanderr: That's probably it, thanks!

server/src/inmantals/server.py
sanderr: Ok, that's clear, thanks!

src/commands.ts
```diff
@@ -32,6 +34,18 @@ export class InmantaCommands {
 		this.context.subscriptions.push(disposable);
 	}
 
+	registerCommands(languageServer: LanguageServer): void {
+		// We have to register these commands each time a diffent language server is being activated or "focussed".
```
sanderr, line 38:
```suggestion
		// We have to register these commands each time a different language server is being activated or "focused".
```

sanderr (approved):

PR #1099, opened 2023-04-11
Issue 1058 tie ls

### LINKED ISSUE #1058: Update LS of vscode extension if newer version available
~- if a newer version of inmanals is available propose the user to install it with a pop-up~
~- add a step for this to the walkthrough.~

Discussed in planning meeting: instead of the above do the following:

- [x] tie the extension to a specific language server version, update the release process to reflect this
- [x] when the extension starts, make sure to update the language server to the tied version (similar to how the language server is installed for the very first time).
- [x] make sure development is still possible, e.g. skip update when inmantals is installed in editable mode

FloLey:
This PR makes the following changes:

- [x] tie the extension to a specific language server version, update the release process to reflect this
- [x] when the extension starts, make sure to update the language server to the tied version (similar to how the language server is installed for the very first time).
- [x] make sure development is still possible, e.g. skip update when inmantals is installed in editable mode

to achieve those 3 points following behaviours where implemented:

1. If the INMANTA_LS_PATH env variable is set, the LS will be installed in editable mode from there.
2. If it is not set the LS will be installed using the version specified in the requirements.txt file packaged with the extension.
3. If no requirements.txt is found the latest version of the LS will be installed
4. The release pipeline now  takes a parameter as input to specify the version of the LS we want to tie to the extension. This version will be written to a requirements.txt file that is packaged with the extension.
5. when the extension starts it checks if the installed LS version is correct, if not it proposes to install the right one.

closes #1058


package.json
```diff
@@ -159,6 +159,7 @@
     "extensionDependencies": [
         "ms-python.python"
     ],
+    "activationEvents": [],
```
FloLey, line 162: this shouldn't be needed but vsce complains if I don't add it :(

.ci/Jenkinsfile-extension-vsce
```diff
@@ -7,6 +7,11 @@ pipeline {
             choices: ['none', 'patch', 'minor', 'major'],
             description: 'Indicates what kind of version bump should be done for the extension.',
         )
+        string(
+            name: "LS_version",
+            defaultValue: "latest",
+            description: "The version(semvar format ex. 1.3.0 or latest) of the LS that will be tied to this release of the extenstion.",
```
Hugo-Inmanta, line 13:
```suggestion
            description: "The version (semver format ex. 1.3.0 or latest) of the LS that will be tied to this release of the extenstion.",
```

.ci/Jenkinsfile-release-all
```diff
@@ -22,6 +22,11 @@ pipeline {
             choices: ['none', 'patch', 'minor', 'major'],
             description: 'Indicates what kind of version bump should be done for the extension. Ignored if release_extension == false',
         )
+        string(
+            name: "server_version",
+            defaultValue: "latest",
+            description: "The version(semvar format ex. 1.3.0 or latest) of the LS that will be tied to this release of the extenstion.",
```
Hugo-Inmanta, line 28:
```suggestion
            description: "The version (semver format ex. 1.3.0 or latest) of the LS that will be tied to this release of the extenstion.",
```

.ci/Jenkinsfile-extension-vsce
```diff
@@ -168,10 +175,21 @@ pipeline {
...
+                        # create the requirements.txt file
+                        if [[ "${LS_version}" == "latest" ]]; then
+                            ${WORKSPACE}/env/bin/pip install inmantals
+                        else
+                            ${WORKSPACE}/env/bin/pip install inmantals=="${LS_version}"
+                        fi
+                        ls_version=$(${WORKSPACE}/env/bin/pip show inmantals | grep Version | awk '{print $2}')
+                        echo "inmantals==${ls_version}" > requirements.txt
+
+                        node_modules/.bin/vsce package
```
Hugo-Inmanta, line 192: Why change from `$(npm bin)/vsce package` to this?

Hugo-Inmanta (approved):
Looks good to me overall :+1:

src/language_server.ts
```diff
@@ -209,11 +299,14 @@ export class LanguageServer {
...
 	 * If the user declines to install the server, returns a Promise that rejects with an error message.
 	 */
-	async proposeInstallLS(diagnoseId: string) {
+	async proposeInstallLS(diagnoseId: string, reason: LanguageServerDiagnoseResult) {
 		if (!this.pythonPath || !fileOrDirectoryExists(this.pythonPath)) {
 			await this.selectInterpreter(diagnoseId);
 		}
-		const response = await window.showErrorMessage(`Inmanta Language Server not installed, run "${this.pythonPath} -m pip install inmantals" ?`, 'Yes', 'No');
+		const msg = reason === LanguageServerDiagnoseResult.wrongLanguageServer
+		? "Wrong version of the Inmanta Language Server installed. Do you want to update? "
```
sanderr, line 307:
```suggestion
		? "A new version of the Inmanta Language Server is available. Do you want to update? "
```
Mostly a suggestion I don't feel too strongly about. The only actual change request here is to not use "wrong version".

Side note: would be cool if we could get some users to use this flow for #1067, not sure how feasible that would be.

src/language_server.ts
```diff
@@ -259,6 +347,7 @@ export class LanguageServer {
 			};
 			return Promise.reject("failed to install LS");
 		} else{
+			log("LS installed with: "+ cmdArgs.join(' '));
```
sanderr, line 350: I like the additional logging but perhaps it would be better to log this right before running the command? This would cover the failure case as well.

src/language_server.ts
```diff
@@ -78,6 +83,85 @@ export class LanguageServer {
...
+	}
+
+	/**
+	 * Checks if the correct version of the Inmanta Language Server is installed.
+	 *
+	 * @returns {boolean} True if the correct version of the Inmanta Language Server is installed, false otherwise.
+	 */
+	isCorrectInmantaLSVersionInstalled(): boolean {
+		// The LS is installed in editable mode via the env var
+		if (process.env.INMANTA_LS_PATH) {
```
sanderr, line 137: I would go one step further here. The env var suffices for our ci but it's not very user friendly for users using editable mode (mostly us when developing). We can use something like `pip list --editable` to check if it is installed in editable mode. If it is we never propose to upgrade. Wdyt?

src/language_server.ts
```diff
@@ -78,6 +83,85 @@ export class LanguageServer {
...
+			return true;
+		}
+		// No requirements specified
+		if (!fs.existsSync(REQUIREMENTS_PATH)) {
```
sanderr, line 141: Just as a sanity check, this would never occur in practice, would it?

src/language_server.ts
```diff
@@ -78,6 +83,85 @@ export class LanguageServer {
...
+			return true;
+		}
+		// Get the expected version from requirement.txt
+		let expectedVersion = null;
+		const requirementTxtContent = fs.readFileSync(REQUIREMENTS_PATH, "utf-8");
+		const inmantaLSPattern = /^inmantals[=<>].*$/gm;
+		const inmantaLSLine = requirementTxtContent.match(inmantaLSPattern)?.[0];
+		if (inmantaLSLine) {
+			expectedVersion = inmantaLSLine.split(/[<=>]{2}/)[1];
```
sanderr, line 150: This only checks for the boundary version. This will suffice for `==` but not for other constraints. So either we should only support `==`, or we should make this a bit more robust. I think it would be nice if we could support `~=` if feasible (see my other comment), otherwise `==` will suffice. I would not make an effort to support the constraints we're not going to use. When we generate the file we can add a comment to it that mentions its purpose and that it only supports `~=`/`==`.

src/language_server.ts
```diff
@@ -13,10 +15,13 @@ import { fileOrDirectoryExists, log } from './utils';
 import { v4 as uuidv4 } from 'uuid';
 import { getLanguageMap } from './extension';
 
-enum LanguageServerDiagnoseResult {
+const REQUIREMENTS_PATH = path.join(__dirname, "..", "requirements.txt");
```
sanderr, line 18: Do you know if this is the same in packaged form?

src/language_server.ts
```diff
@@ -78,6 +83,85 @@ export class LanguageServer {
...
+	}
+
+	/**
+	 * Returns the version of the installed Inmanta Language Server, or null if it's not installed.
+	 *
+	 * @returns {string | null} The version of the installed Inmanta Language Server, or null if it's not installed.
+	 */
+	getInstalledInmantaLSVersion(): string | null {
+		try {
+		  const output = cp.execSync(`${this.pythonPath} -m pip show inmantals`).toString();
```
sanderr, line 121: Same remark about using `pip freeze` here.

sanderr (changes requested):

src/language_server.ts
FloLey: It is easy :p just install the wrong version of inmantals in your venv by hand and start the extension. I've tried it using the .vsix file.

src/language_server.ts
FloLey: in production it should never happen. But in dev there is no requirements.txt file as it will only be added by the ci/cd.

src/language_server.ts
FloLey: I tested it by installing the .vsix produced by jenkins and it worked for me

.ci/Jenkinsfile-extension-vsce
```diff
@@ -168,10 +175,21 @@ pipeline {
...
-                        $(npm bin)/vsce package
+
+                        # create the requirements.txt file
+                        if [[ "${LS_version}" == "latest" ]]; then
+                            ${WORKSPACE}/env/bin/pip install inmantals
+                        else
+                            ${WORKSPACE}/env/bin/pip install inmantals=="${LS_version}"
+                        fi
+                        ls_version=$(${WORKSPACE}/env/bin/pip show inmantals | grep Version | awk '{print $2}')
+                        echo "inmantals==${ls_version}" > requirements.txt
```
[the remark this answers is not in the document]
FloLey: so just changing `==` by `~=` right?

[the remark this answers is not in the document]
sanderr: Indeed.

src/language_server.ts
FloLey: By supporting `~=`, do you mean that it the user has version "1.2.3" of the LS, and that version "1.2.4" is released, he should also receive the pop-up to install the new version? because this I currently not support at all. I never check what is available on the pip index. Should I try that?

sanderr: That would be even better but I thought that might bring us too far. What I had in mind was that if the constraint is `~=1.2.3` and the user has `1.2.4` that we wouldn't keep asking them to upgrade.

FloLey: Ok i'll try to add support for what you propose. This should be feasible with not to much effort.

src/language_server.ts
```diff
@@ -143,12 +154,14 @@ export class LanguageServer {
...
-		const inmantaLSPattern = /^inmantals[=<>].*$/gm;
-		const inmantaLSLine = requirementTxtContent.match(inmantaLSPattern)?.[0];
+		const inmantaLSPattern = /^inmantals(==|~=).*$/gm;
+		const inmantaLSLine = requirementTxtContent.match(inmantaLSPattern)[0];
 		if (inmantaLSLine) {
-			expectedVersion = inmantaLSLine.split(/[<=>]{2}/)[1];
-		}
+			operator = inmantaLSLine.match(/(==|~=)/)?.[0] ?? "==";
+			expectedVersion = inmantaLSLine.split(/(==|~=)/)[1];
+		  }
```
sanderr, line 164:
```suggestion
		}
```

src/language_server.ts
sanderr: Right.

.ci/Jenkinsfile-extension-vsce
```diff
@@ -168,10 +175,21 @@ pipeline {
                     sh '''
                         # install the extension for its vsce dependency, required for building and publishing the extension
                         # and populate vscodeignore with exceptions to node_modules of the needed modules.
+                        # add the requirements.txt file to tie the extenstion to a server version
```
sanderr, line 178: Your comment [here](https://github.com/inmanta/vscode-inmanta/pull/1099#discussion_r1166518902) got me thinking about when we add this. It would be more in line with our current release process to add this in the "Setup next branch for major or minor release" stage so it is committed to the next (and after that the stable) branch.

sanderr (changes requested):
Nice work with the `~=`, it came out simpler than I imagined and I think it's a solid improvement.

.ci/Jenkinsfile-extension-vsce
FloLey: i'll have a look :p

FloLey: I made the change but I did not retest the whole flow as I expect it to behave in the same way but I would be harder for me to get the .vsix.

.ci/Jenkinsfile-extension-vsce
```diff
@@ -92,6 +92,15 @@ pipeline {
...
                             git commit -m "set release date: ${release_date}" CHANGELOG.md
 
+                            # add a requirements.txt file to tie the extenstion to a server version
+                            if [[ "${LS_version}" == "latest" ]]; then
+                                ${WORKSPACE}/env/bin/pip install inmantals
+                            else
+                                ${WORKSPACE}/env/bin/pip install inmantals=="${LS_version}"
+                            fi
+                            ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep inmantals | cut -d'=' -f3
+                            echo "inmantals~=${ls_version}" > requirements.txt
```
sanderr, line 102: Two remarks (I should probably have given a short introduction as to how this pipeline works because it's quite niche):
- you'll need to commit this
- I don't think we want this on the `issue/version-bump-extension` branch (which gets merged both into master and next eventually) but only on the `next` branch, so you'll need to move this down a bit.

sanderr (changes requested):

sanderr (approved):

PR #1116, opened 2023-04-19
Fix issue install loop and images

FloLey:
fix an issue where installing the right LS version would still prompt to install it again and fix issue where images where not packaged


src/language_server.ts
```diff
@@ -160,7 +160,7 @@ export class LanguageServer {
 		const inmantaLSLine = requirementTxtContent.match(inmantaLSPattern)[0];
 		if (inmantaLSLine) {
 			operator = inmantaLSLine.match(/(==|~=)/)?.[0] ?? "==";
-			expectedVersion = inmantaLSLine.split(/(==|~=)/)[1];
+			expectedVersion = inmantaLSLine.split(/(==|~=)/)[2];
```
FloLey, line 163: I don't know how this happened. :grimacing:

PR #1129, opened 2023-04-26
Issue/shadow release fix v1 module code navigation

Hugo-Inmanta:
When navigating a v1 module, ctrl-clicking would navigate to the module in the libs folder of the tmp project, instead of navigating to the opened module.


wouterdb (approved):
I assume this was tested by hand?

Hugo-Inmanta:
> I assume this was tested by hand?

Yes!

PR #1147, opened 2023-05-04
Issue/1128 duplicate hover

FloLey:
when hovering over a plugin and holding the 'ctrl' button to navigate, 2 doc string was shown twice.
This pr should fix that.
Peek 2023-05-04 15-12


FloLey:
I first tried to fix the plugin ranges in core but I didn't find another way than to uses regexes and I didn't feel that was a good idea in the compiler (opening a file and apply the regex to each plugin definition).

sanderr (approved):

Hugo-Inmanta (approved):

PR #1154, opened 2023-05-08
Issue/1134 stability issues env selector

### LINKED ISSUE #1134: stability issues env selector
There are some stability issues with the env selector:

1. when opening an inmanta module on a fresh vscode and select a venv, it is not recognized by the bottom bar. It keeps showing "Select Interpreter" in orange. (needs to open a .cf file quickly) 

2. when opening a Python file, it can triggers a second venv box to appear in the menu (the one from the python extension). This one does show the correct venv and It can be changed. Ours remains stale.

3. close and reopen vscode to work around the issue described in 1. I open a Python file, once again highlighting issue 2. We can change the venv from there (behavior is the same regardless of which of the two boxes I use). This works fine but our box disappears. If we then open an inmanta file again, no box remains and the only way to get it/them back is to restart vscode.

4. select venv, allow it to install inmantals, restart vscode to make sure no lingering side effects of the first 3 issues remain. Then select different venv, uninstall inmantals from the previous one in a terminal, allow the extension to install inmantals in the new one. Observe that inmantals is installed in the old one and not in the new one. I assume the LS is also running from this old (wrong) venv.

5. If no venv is selected at starteup (clean install), the default venv will be selected.

6. this error was reported when switching venvs:
```
2023-04-21 14:31:20.656 [error] Error: shutdown() got an unexpected keyword argument 'cancel_futures'
	at handleResponse (/home/wouter/.vscode/extensions/inmanta.inmanta-1.3.0/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)
	at handleMessage (/home/wouter/.vscode/extensions/inmanta.inmanta-1.3.0/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)
	at processMessageQueue (/home/wouter/.vscode/extensions/inmanta.inmanta-1.3.0/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)
	at Immediate.<anonymous> (/home/wouter/.vscode/extensions/inmanta.inmanta-1.3.0/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)
	at process.processImmediate (node:internal/timers:466:21)
```
=> cancel_futures was introducted in python3.9

Hugo-Inmanta:
closes #1134

Fixes visibility issues of the python env selector (no more duplication or disapearance).
Fixes issues when selecting an interpretor with no ls installed in it.
Fixes the issue where navigation would only work for the top most _init.cf file when working on a module

~~Not super satisfied with the approach to globally set infoVisibility to "never" (in python_extension.ts ) it feels a bit hackish and I try to set it back to its old value on deactivation but it doesn't seem to work perfectly.~~

=> fixed this using the old approach again


src/python_extension.ts
```diff
@@ -23,8 +24,27 @@ export class PythonExtension {
 		this.executionDetails = pythonApi.settings.getExecutionDetails(workspace.workspaceFolders?.[0].uri);
 		this.pythonApi = pythonApi;
 		this.onChange(pythonApi);
+
+	}
+
+	async hidePythonButtonCfg(): Promise<void> {
```
FloLey, line 30: I guess this works but I think it would prefer that our extension only fiddles with our own stuff and configs: So hide our button whenever the python one is visible. Did we had the case where the 2 where shown when a .cf file was open our only with .py files?

src/python_extension.ts
```diff
@@ -56,6 +76,7 @@ export class PythonExtension {
 
 
 	async updatePythonVersion(): Promise<string> {
+		console.log("updatePythonVersion");
```
FloLey, line 79:
```suggestion

```

src/python_extension.ts
```diff
@@ -73,22 +94,27 @@ export class PythonExtension {
...
-		if (document) {
-			const uri = document.uri;
-			let folder = workspace.getWorkspaceFolder(uri);
-			if (!folder) {
-				// not clicking on a cf file -> do nothing
-				return;
+		if (documentURI) {
+			try{
+				let folder = workspace.getWorkspaceFolder(documentURI);
+				console.log("folder", folder);
```
FloLey, line 107:
```suggestion
```

src/python_extension.ts
```diff
@@ -106,21 +132,22 @@ export class PythonExtension {
...
 		} else {
+			this.restorePythonCfg();
 			this.inmantaEnvSelector.hide();
 		}
 	}
 
 
 	addEnvSelector():void {
 		// Add the EnvSelectorWindow
+		console.log("Adding env selector");
```
FloLey, line 147: use log() or clean up

server/src/inmantals/server.py
```diff
@@ -241,8 +241,27 @@ def create_tmp_project(self) -> str:
...
+                if os.path.isdir(sub_path):
+                    files.extend(_get_name_spaces(sub_path, prefix + "::" + entry))
+
+            return files
+
+        name_spaces = _get_name_spaces(os.path.join(self.folder_path, "model"), module_name)
         with open(os.path.join(inmanta_project_dir, "main.cf"), "w+") as fd:
-            fd.write(f"import {module_name}\n")
+            for name in name_spaces:
+                fd.write(f"import {name}\n")
```
FloLey, line 264: Why did we need this?

FloLey:
merging master into your branch should fix the issue about '`SyntaxError: future feature annotations is not defined`'

server/src/inmantals/server.py
Hugo-Inmanta: We need to import all sub-modules, otherwise they won't be know to the project when performing the anchormap computation, (and so navigation wouldn't work for them). I realized this when testing on the big junos-qfx module, navigation was only working on the .cf file living at the root of the folder

src/python_extension.ts
Hugo-Inmanta: > it would prefer that our extension only fiddles with our own stuff and configs

I agree, I will try something cleaner, and iirc it was only when opening .py files

sanderr:
@Hugo-Inmanta I would think so, it's [part of the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references). And if it's giving performance issues it must be, right?

Hugo-Inmanta:
It's not the method call itself that's causing issues, its the computation of the reverse_anchormap (that's allegedly used in this method, but I'm questioning that's it is ever called)

src/extension.ts
```diff
@@ -56,22 +57,32 @@ export async function activate(context: ExtensionContext) {
...
 		inmantaCommands.registerCommands(languageServer);
 	}
 
 	async function didOpenTextDocument(document: TextDocument): Promise<void> {
+		// We are only interested in .cf files
+		if (document.languageId !== 'inmanta' || (document.uri.scheme !== 'file')) {
+			return;
+		}
+		
+		pythonExtensionInstance.updateInmantaEnvVisibility(document.uri)
```
sanderr: I'm not entirely sure how to interpret this. What would happen if you have an inmanta file open and then open a Python file or the `requirements.txt`?

src/python_extension.ts
sanderr: If I understand correctly this may invalidate the trouble you were still having with restoring the Python button so I'll hold off on looking into that for now.

sanderr (changes requested):
Sorry it took me so long to review this. Team week and all.

server/src/inmantals/server.py
```diff
@@ -241,8 +241,27 @@ def create_tmp_project(self) -> str:
...
+            files: List[str] = []
+            init_cf = os.path.join(curdir, "_init.cf")
+            if not os.path.exists(init_cf):
+                return files
+
+            files.append(prefix)
+            for entry in os.listdir(curdir):
+                sub_path = os.path.join(curdir, entry)
+                if os.path.isdir(sub_path):
+                    files.extend(_get_name_spaces(sub_path, prefix + "::" + entry))
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, this means that the performance issue is even worse than what I had (since I was missing some imports)

src/extension.ts
```diff
@@ -23,21 +23,22 @@ export var languageServers: Map<string, LanguageServer> = new Map();
 let sortedWorkspaceFolders: string[] | undefined;
 workspace.onDidChangeWorkspaceFolders(() => sortedWorkspaceFolders = undefined);
 
-
+var pythonExtensionInstance ;
```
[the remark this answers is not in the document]
Hugo-Inmanta: Fixed

server/src/inmantals/server.py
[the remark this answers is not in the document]
Hugo-Inmanta: Should be fixed now

server/src/inmantals/server.py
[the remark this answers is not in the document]
Hugo-Inmanta: > Good catch. This wasn't a reported issue, was it? 
Indeed it was not!

As discussed, I went with the former approach of importing all submodules in the tmp_dir's main.cf.
As suggested I tested on the vcenter module, no performance issue as far as I can tell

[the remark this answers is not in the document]
sanderr: How many seconds from opening/saving a file to end of compile?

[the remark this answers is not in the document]
Hugo-Inmanta: Pretty quick:
2023-05-17 10:00:29,723 inmantals.server         DEBUG   Called: textDocument_didSave 
2023-05-17 10:00:29,884 inmantals.server         INFO    Compilation succeeded

sanderr (changes requested):
I played with it just now and I have a few observations:
- when you are working on a project/module, then open a new module for the very first time (delete `~/.config/Code`) the venv selector is very slow to appear
- in this same scenario, selecting the venv once it does appear does not affect the contents of the box. Switching to a Python file seems to refresh it. Switching back restores the old invalid value. The compiles do seem to use the appropriate venv
- not related to this PR perhaps but I noticed it while testing this: updating the repos in the extensions' settings does not seem to take effect: the project I got still had the wrong repo.
- the `repo` setting in the project.yml is not written as a valid yaml dict

Hugo-Inmanta:
> the repo setting in the project.yml is not written as a valid yaml dict

It is a list of dicts, I think it's valid ?

src/extension.ts
sanderr: I get another venv box with another venv selected when I do this.

Hugo-Inmanta: What's your setting value for "python.interpreter.infoVisibility" ?

sanderr:
> It is a list of dicts, I think it's valid ?

It's though to interpret, there are a lot of quotes, it might just be a string? I deleted it before thinking to make a copy. The `inmanta project install` fails on it anyway.

Hugo-Inmanta:
> not related to this PR perhaps but I noticed it while testing this: updating the repos in the extensions' settings does not seem to take effect: the project I got still had the wrong repo.

Works on my machine:
[Screencast from 17-05-2023 10:46:46.webm](https://github.com/inmanta/vscode-inmanta/assets/98876549/ef9a7e94-bdbd-4149-a02a-ed5e76d50af4)

src/extension.ts
sanderr: `onPythonRelated`

Hugo-Inmanta: This is the behavior I have: 

[Screencast from 17-05-2023 10:53:05.webm](https://github.com/inmanta/vscode-inmanta/assets/98876549/e0e6db74-dda7-4a4a-b6f4-77338319938f)

Hugo-Inmanta: Did you install the ls in editable mode ?

src/extension.ts
```diff
@@ -23,21 +23,22 @@ export var languageServers: Map<string, LanguageServer> = new Map();
...
 	}
 
 	log("Activate Python extension");
 	await pythonExtension.activate();
-
 	// Start a new instance of the python extension
-	let pythonExtensionInstance = new PythonExtension(pythonExtension.exports);
+	pythonExtensionInstance = new PythonExtension(pythonExtension.exports);
 	//add the EnvSelector button
+	await pythonExtensionInstance.hidePythonButtonCfg();
```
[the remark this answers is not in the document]
sanderr: This comment is still valid, right?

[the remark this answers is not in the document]
Hugo-Inmanta: Indeed :+1: 
Done now

FloLey (approved):
cool I like it a lot more :D

sanderr (approved):

PR #1179, opened 2023-05-17
Issue/1133 investigate crash when running a v1 module in iso4

Hugo-Inmanta:
Closes https://github.com/inmanta/vscode-inmanta/issues/1133


server/src/inmantals/server.py
```diff
@@ -196,22 +196,34 @@ def create_tmp_project(self) -> str:
...
+                    mv2 = module.ModuleV2(project=None, path=self.folder_path, is_editable_install=True)
+                    module_name = mv2.name
+                    self.kind = module.ModuleV2
+            return module_name
 
-        elif os.path.exists(v2_metadata_file):
-            mv2 = module.ModuleV2(project=None, path=self.folder_path, is_editable_install=True)
-            module_name = mv2.name
-            self.kind = module.ModuleV2
+        module_name = _get_module_name()
```
Hugo-Inmanta, line 226: The `_get_module_name` method fixes a bug when using iso4 (module.ModuleV2 doesn't exist for core versions < 5)
- it keeps the same behavior as before for core versions >= 5
- for core versions < 5 we only check if a V1 module is open

server/src/inmantals/server.py
```diff
@@ -553,7 +565,10 @@ async def shutdown(self, **kwargs):
         logger.debug("shutdown requested...")
         if self.tmp_project:
             self.tmp_project.cleanup()
-        self.threadpool.shutdown(cancel_futures=True)
+        if CORE_VERSION < version.Version("5.dev"):
+            self.threadpool.shutdown()
+        else:
+            self.threadpool.shutdown(cancel_futures=True)
```
Hugo-Inmanta, line 571: The cancel_futures argument doesn't exist for python < 3.9 https://docs.python.org/3.6/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

server/src/inmantals/server.py
```diff
@@ -682,7 +697,11 @@ async def textDocument_hover(self, textDocument, position):
         data = list(range)[0].data
         docstring = textwrap.dedent(data.docstring.strip("\n")) if data.docstring else ""
         docstring = docstring.replace(" ", "&nbsp;").replace("\n", "\n\n").strip()
-        definition = self.get_definition(data).strip()
+        try:
```
Hugo-Inmanta, line 700: Fixes a bug when hovering over the `Entity` keyword e.g. in 
```
implementation none for Entity:
    """
        An empty implementation that can be used as a safe default.
    """
end
```
As I understand it this keyword is defined in a dummy file called "internal" that isn't meant to be navigated to ?

server/src/inmantals/server.py
```diff
@@ -629,7 +644,7 @@ def convert_location(self, location: Location):
     def get_definition(self, target: AnchorTarget) -> str:
         file_path = target.location.file
         start_line = target.location.lnr - 1
-        with open(file_path, "r") as f:
+        with open(file_path, "r", encoding="utf8") as f:
```
Hugo-Inmanta, line 647: Fixes a bug when hovering e.g. over `std::none` in iso4. When encoding is not specified, locale.getpreferredencoding(False) is called to get the current locale encoding and it resolves to `ANSI_X3.4-1968` on my system and it throws this error:
`  Message: 'ascii' codec can't decode byte 0xc2 in position 1113: ordinal not in range(128)`

server/src/inmantals/server.py
```diff
@@ -278,7 +290,7 @@ def install_project(self, attach_cf_cache: bool):
         module.Project.set(module.Project(self.inmanta_project_dir, attach_cf_cache=attach_cf_cache))
         env_path = module.Project.get().virtualenv.env_path
         logger.info("Installing project at %s in env %s.", self.inmanta_project_dir, env_path)
-        if self.kind == module.ModuleV2:
+        if CORE_VERSION >= version.Version("5.dev") and self.kind == module.ModuleV2:
```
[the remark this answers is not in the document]
Hugo-Inmanta: Nice catch, I didn't realize this was only called for iso6

sanderr (approved):

PR #1215, opened 2023-06-07
fix issues with the release pipeline for vscode

FloLey:
The are some bugs in the release pipeline. This PR fixes those bugs
Also add an additional check to stop the release if no inmantals version/package is found when building the vs-code extension

### DIFF
--- a/.ci/Jenkinsfile-extension-vsce
+++ b/.ci/Jenkinsfile-extension-vsce
@@ -107,6 +107,12 @@ pipeline {
                                 ${WORKSPACE}/env/bin/pip install inmantals=="${LS_version}"
                             fi
                             ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep inmantals | cut -d'=' -f3)
+
+                            if [[ -z "${ls_version}" ]]; then
+                                echo "Error: ls_version not found. Stopping execution."
+                                exit 1
+                            fi
+
                             echo "inmantals~=${ls_version}" > requirements.txt
                             git add requirements.txt -f
                             git commit -m "add requirements.txt file to tie the inmantals version (${ls_version}) to the extension"
--- a/.ci/Jenkinsfile-extension-vsce
+++ b/.ci/Jenkinsfile-extension-vsce
@@ -100,7 +100,7 @@ pipeline {
                             git merge -X theirs issue/version-bump-extension
 
 
-                            if [ -f "${WORKSPACE}/requirements.txt" ]; then
+                            if [ -f "requirements.txt" ]; then
                                 echo "Error: requirements.txt already exists. Stopping execution."
                                 exit 1
                             fi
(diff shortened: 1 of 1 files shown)


Hugo-Inmanta (approved):
Looks good to me :+1:

.ci/Jenkinsfile-extension-vsce
```diff
@@ -106,9 +106,17 @@ pipeline {
...
+                            ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep inmantals | cut -d'=' -f3)
+
+                            if [[ -z "${ls_version}" ]]; then
+                                echo "Error: ls_version not found. Stopping execution."
+                                exit 1
+                            fi
+
                             echo "inmantals~=${ls_version}" > requirements.txt
-                            git commit -m "add requirements.txt file to tie the inmantals version (${ls_version}) to the extension" requirements.txt
+                            # force add requirements.txt as it is part of .gitignore
```
arnaudsjs, line 117: The assumption here is that we don't have a `requirements.txt` file and we add one. If not we risk that we override the content of the existing requirements.txt file. Shouldn't we make this explicit and add a check that verifies that no requirements.txt exists.

.ci/Jenkinsfile-extension-vsce
```diff
@@ -106,9 +106,17 @@ pipeline {
                             else
                                 ${WORKSPACE}/env/bin/pip install inmantals=="${LS_version}"
                             fi
-                            ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep inmantals | cut -d'=' -f3
+                            ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep inmantals | cut -d'=' -f3)
```
arnaudsjs, line 109: I would make the grep more specific. Otherwise we might accidentally match an invalid line and that would be disaster if we push that into the release.

```suggestion
                            ls_version=$(${WORKSPACE}/env/bin/pip freeze | grep -E '^inmantals==[0-9]+\.[0-9]+\.[0-9]+$' | cut -d'=' -f3)
```

arnaudsjs (changes requested):

.ci/Jenkinsfile-extension-vsce
FloLey: currently there should never be a requirements.txt. It is onlu used by the extention to install the right version of the LS. 
But I'm not oposed to check that to and to stop the pipeline if there is already one. I'll add the check

.ci/Jenkinsfile-extension-vsce
```diff
@@ -100,15 +100,29 @@ pipeline {
                             git merge -X theirs issue/version-bump-extension
 
 
+                            if [ -f "${WORKSPACE}/requirements.txt" ]; then
```
arnaudsjs, line 103: This is not correct, is it. We are in the sub-directory `vscode-inmanta`. Checking at `vscode-inmanta` would mean that we check for the file outside of the vscode-inmanta repository.
```suggestion
                            if [ -f "requirements.txt" ]; then
```

FloLey: oups indeed, sorry and thanks! I went to fast

arnaudsjs (approved):

PR #1513, opened 2023-12-12
Fix bug in tox file

arnaudsjs:
This PR fixes a bug in the `tox.ini` file that caused the tests to run against an incorrect version of inmanta-core.


FloLey:
I added iso7-stable to the PR (part of post release)

FloLey (approved):

Hugo-Inmanta (approved):

PR #1514, opened 2023-12-12
Add workaround for bug in pythonExtension.getExecutionDetails() method

### LINKED ISSUE #1508: new python extension release breaks our tests
Version [2023.22.0](https://github.com/microsoft/vscode-python/releases/tag/v2023.22.0) of the Python extension was released yesterday. Our tests are hanging since. Constraining the version in `runTest.ts` to the previous version results in success.

The cause of the issue should be investigated and addressed.

arnaudsjs:
[This bug](https://github.com/microsoft/vscode-python/issues/22617) in the Python extension causes the `this.pythonApi.settings.getExecutionDetails().execCommand[0]` to return the path to the root of the venv instead of the path to the python binary in that venv. This PR applies a workaround for this issue until it's fix in the python extension.

Closes #1508

### DIFF
--- a/src/test/runTest.ts
+++ b/src/test/runTest.ts
@@ -40,7 +40,7 @@ async function main() {
 
 		const vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
 		const cliPath = resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath, "linux-x64");
-		cp.spawnSync(cliPath, ['--install-extension', 'ms-python.python'], {
+		cp.spawnSync(cliPath, ['--install-extension', 'ms-python.python', '--force'], {
 		encoding: 'utf-8',
 		stdio: 'inherit'
 		});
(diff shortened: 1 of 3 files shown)


src/test/runTest.ts
```diff
@@ -40,7 +40,7 @@ async function main() {
 
 		const vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
 		const cliPath = resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath, "linux-x64");
-		cp.spawnSync(cliPath, ['--install-extension', 'ms-python.python'], {
+		cp.spawnSync(cliPath, ['--install-extension', 'ms-python.python', '--force'], {
```
FloLey, line 43: It doesn't work without this anymore?

FloLey (approved):

src/test/runTest.ts
arnaudsjs: This flag makes sure that we always check whether there is a newer version of the extensions available and upgrade to it if there is one. It should just make sure that we get faster feedback when there is a breaking change. So to answer your question: without this flag the tests should also work.

Hugo-Inmanta (approved):

PR #1650, opened 2024-03-20
Fix language servers stop/start logic + fix memory leak

arnaudsjs:
changes:

* Fix bug that could cause multiple instances of the language server to be running simultaneously.
* Fix memory leak by resetting the `inmanta.plugins.PluginMeta` and `inmanta.export.Exporter` classes before every compile.

### DIFF
--- a/src/language_server.ts
+++ b/src/language_server.ts
@@ -499,10 +499,14 @@ export class LanguageServer {
 
 	/**
 	 * Starts the TCP server and the Inmanta Language Server.
+	 * This function should always run under the `mutex` lock.
 	 *
 	 * @param clientOptions The options for the language client.
 	 */
 	private async startTcp(clientOptions: LanguageClientOptions) {
+		if(this.client){
+			return;
+		}
 		const host = "127.0.0.1";
 		// Get a random free port on 127.0.0.1
 		const serverPort = await getPort({ host: host });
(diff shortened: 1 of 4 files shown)


FloLey (approved):

src/language_server.ts
```diff
@@ -619,33 +627,46 @@ export class LanguageServer {
...
+				await this._stopServerAndClient();
+			});
+		} else {
+			await this._stopServerAndClient();
+		}
+	}
+
+	/* 
+	 * Stops the language server and its client.
+	 * This function should always run under the `mutex` lock.
```
sanderr, line 655: This says "always", yet `stopServerAndClient` has a boolean to control it.

sanderr (changes requested):

src/language_server.ts
arnaudsjs: This docstring belongs to the `_stopServerAndClient()` method and not to the `stopServerAndClient` method

sanderr: Yes, I meant to say that `stopServerAndClient`'s boolean controls whether or not it calls `_stopServerAndClient` under the mutex lock. So it violates the docstring of this method. Either the docstring is too strict, or the boolean should not be there in the caller method.

sanderr: Or should the boolean be false only if it is already running under the lock?

arnaudsjs: yes

sanderr (approved):

PR #1953, opened 2024-10-04
Add support for ISO version that are not using pkg_resources internally

hlloreda:
In the latest dev release, we have dropped pkg_resources partially. The `are_installed` method is not expecting to receive `pkg_resources.Requirement` instances

### DIFF
--- a/server/tests/test_smoke.py
+++ b/server/tests/test_smoke.py
@@ -263,12 +263,13 @@ async def test_working_on_v1_modules(client, caplog):
     with open(req_file, "r") as f:
         reqs = list(parse_requirements(f.readlines()))
 
+    str_reqs = [str(e) for e in reqs]
     venv = env.VirtualEnv(env_path)
     venv.use_virtual_env()
 
     assert Requirement.parse("inmanta-module-dummy>=3.0.8") in reqs
     # This dummy module is used because it is only present in the dev artifacts and not in pypi index.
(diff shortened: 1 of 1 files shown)


server/tests/test_smoke.py
```diff
@@ -263,12 +263,13 @@ async def test_working_on_v1_modules(client, caplog):
     with open(req_file, "r") as f:
         reqs = list(parse_requirements(f.readlines()))
 
+    str_reqs = [str(e) for e in reqs]
     venv = env.VirtualEnv(env_path)
     venv.use_virtual_env()
 
     assert Requirement.parse("inmanta-module-dummy>=3.0.8") in reqs
```
hlloreda, line 270: We cannot just drop pkg_resources yet (until a new stable version has been released)

Hugo-Inmanta: I think I'm missing some context here, but doesn't the fact that the test fails means that people will encounter that same issue when opening a v1 module in vscode ?

Hugo-Inmanta (commented):
See comment

server/tests/test_smoke.py
hlloreda: I'm not sure if this will break the VScode experience, so I'm asking for your review @wouterdb

wouterdb (approved):
As this only affects testing, it is fine for me

PR #2142, opened 2025-01-30
POC - improve tests

LukasStordeur:
# Description

- Update the readme with an explanation on how to run the tests/develop for the extention.
- Reformat the code and do a few improvements on overal structure of the extention.ts file. 
- Updated the launch.json to contain the required env-variables for the tests.

### DIFF
--- a/src/python_extension.ts
+++ b/src/python_extension.ts
@@ -37,7 +37,7 @@ export class PythonExtension {
 		return PythonExtension.getPathPythonBinary(this.executionDetails.execCommand[0]);
 	}
 
-	get virtualEnvName(): string | null {
+	get virtualEnvName(): string {
 		return this.pythonPathToEnvName(this.pythonPath);
 	}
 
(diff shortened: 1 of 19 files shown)


src/python_extension.ts
```diff
@@ -105,60 +112,115 @@ export class PythonExtension {
...
+			const folder = workspace.getWorkspaceFolder(documentURI);
+			if (folder) {
+				return this.pythonPathToEnvName(getLanguageMap().get(folder.uri.toString()).pythonPath);
 			}
-			
+		} catch (error) {
+			console.error(`Failed to get virtual environment name: ${error}`);
 		}
+		return undefined;
+	}
```
jptrindade, line 169: Can this have a default value as an argument (this.virtualEnvName) so we can return that instead of undefined

jptrindade (approved):
small comments, but overall looks good to me. Linting is a major improvement, seeing how it was before is scary

Hugo-Inmanta (approved):
LGTM!

PR #2180, opened 2025-02-19
Typescript improvements

LukasStordeur:
# Description

- Refactor the code to use the python extention package instead of an outdated version that was copied into the project.
- Add a step to the setup assistant to suggest creating an venv
- Upgrade the version check for python to require 3.11 and above.
- Add an "inmanta" output channel to display logs when running the extension to allow better understanding of errors for the users. 

> Note to Reviewer: Please try to run the extension and confirm you can run the install flow. I ran it multiple times without issues, but an additional check of someone running a real linux devise can't hurt


wouterdb (approved):

Hugo-Inmanta (approved):
LGTM apart from an old typo, tested the install flow as well :+1:

PR #2282, opened 2025-04-15
Package extension with Webpack + add VSIX test suite

LukasStordeur:
#  description

- Adjust the code to not use the python package from the node_modules, but use the one installed in the vs-code instance. 
- Add support to run tests with a -vsix flag, to test against the package we actually release in the stores, and add a stage step for it in the pipeline.
- Add  webpack configuration to bundle the files in an optimized way.
- Update package.json scripts

VSIX file was tested locally aswell.


src/python.ts
```diff
@@ -70,7 +66,7 @@ export async function initializePython(disposables: Disposable[]): Promise<void>
 
         if (api) {
             disposables.push(
-                api.environments.onDidChangeActiveEnvironmentPath((event) => {
+                api.environments.onDidChangeActiveEnvironmentPath((event: any) => {
```
matborowczyk, line 69: I don't know about the typing there and below but is there a way to remove any?

matborowczyk (changes requested):

src/python.ts
LukasStordeur: I tried to find the exact typing, but my esLinter didn't make me much wiser. Don't know if it's worth the effort to dig deeper on that part right now.

wouterdb (approved):

matborowczyk (approved):

Issue #1954, opened 2024-10-04
Drop pkg_resources


hlloreda:
At that time, we didn't have a stable release that wasn't providing helper functions to parse requirements (without relying on `pkg_resources`), so the dependency could not be removed.

PR #2777, opened 2025-09-17
use importlib instead of pkg_resource

Hugo-Inmanta:
Closes https://github.com/inmanta/vscode-inmanta/issues/1954


Hugo-Inmanta:
Tested that the following deprecation warning is gone:
> /home/hugo/work/inmanta/github-repos/vscode-inmanta/server/src/inmantals/server.py:37: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources

Tested that basic functionalities still work e.g. ctrl-click, hover

arnaudsjs (approved):

